PR #13, opened 2021-01-30
fixed one wrong module name and three wrong class names

FlyNeopolitan:
Already Sorted.


winglam:
LGTM. Thanks for the changes!

PR #25, opened 2021-05-18
Repaired and verified the dataset

IntHelloWorld:
- Verified every line in the dataset.
- Repaired some wrong lines in the original dataset.
- Created a new benchmark.csv in which all of the test method can be rightly found.

Notice that all of the test.java files are take into consideration.


winglam:
> Notice that all of the test.java files are take into consideration.
What does it mean that all test.java files are taken into consideration?

> Verified every line in the dataset
How did you verify every line? Do you mind sharing the code you used as well?

> Created a new benchmark.csv in which all of the test method can be rightly found.
Why not just directly make the changes to pr-data.csv? Were many tests reported in pr-data.csv not found? If you manually grep for the missing tests, would it be possible to find them?

darko-marinov:
@IntHelloWorld do you still have some interest in this repo? If so, please let us know. Meanwhile, I'll close the PR for now.

PR #28, opened 2021-05-29
Fix category 'ND' in pr-data.csv

jereparla:
The test org.apache.dubbo.rpc.protocol.dubbo.MultiThreadTest.testDubboMultiThreadInvoke had as its Category ‘NOD;ND’. ND has been deleted because it is not a valid category.


winglam:
LGTM. Thanks for the change!

PR #41, opened 2021-06-13
Added a new flaky test

mohasarc:
I have noticed that this test is flaky, but it hasn't already been included in the database.


pr-data.csv
```diff
@@ -1876,6 +1876,7 @@ https://github.com/vojtechhabarta/typescript-generator,caef58bfe55a8cd1fc5d7368c
 https://github.com/vojtechhabarta/typescript-generator,caef58bfe55a8cd1fc5d7368c19e9ff818a7b7fa,typescript-generator-core,cz.habarta.typescript.generator.parser.GsonParserTest.test,ID,,,
 https://github.com/vojtechhabarta/typescript-generator,caef58bfe55a8cd1fc5d7368c19e9ff818a7b7fa,typescript-generator-core,cz.habarta.typescript.generator.parser.Jackson2ParserPropertiesTest.testPropertyOrder,ID,,,
 https://github.com/vojtechhabarta/typescript-generator,caef58bfe55a8cd1fc5d7368c19e9ff818a7b7fa,typescript-generator-core,cz.habarta.typescript.generator.ReadOnlyWriteOnlyTest.test,ID,,,
+https://github.com/vojtechhabarta/typescript-generator,107eb2f854692bd1ad466596f62b21fcc3fc735c,typescript-generator-core,cz.habarta.typescript.generator.TaggedUnionsTest.testIntermediateUnions,ID,,,,
```
winglam, line 1879: How did you notice this test being flaky?

Also, the change appears to have an extra comma at the end

mohasarc: I was working on fixing the other tests of the same class, when I noticed this test has the same structure as the other tests but is not included in the IDoFT database, so I tested it with the NonDex tool, and it was actually flaky.
I also have removed the comma.

winglam:
LGTM. Thanks for the change!

PR #42, opened 2021-06-19
Format Checker

jereparla:
## Fixes and changes

- [x] Managed to run it on the remote machine using `python3.7 format_checker/main.py`, using just `python3` did not work, since it resorts to version 3.5.2, and the package `errorhandler` didn't work on that for some reason (how did you make that work?). Anyhow, the issue was in the bash script used for sorting, I had to change the line that was `tail +2` to `tail -n +2`, based on [this](https://stackoverflow.com/a/44054842/9610524). I then misplaced some lines to test it and it seems to be working fine. 
- [x] Fixed the issue with the rules that didn't allow an empty PR Link with a blank status. There was nothing weird about it, just that I implemented something that I shouldn't have.
- [x] Added a .gitignore file (to ignore common undesired files/directories related to Python, macOS, Linux).
- [x] Changed name of the job running the format checker to `format-checker` from `build`.
- [x] Updated README.
- [x] Fixed typo in `ci.yml` ("strin" instead of "string").


winglam (commented):
Just tested these changes on the idem2 machine and they worked quite well. 

One minor suggestion would be to add a rule for checking that if a PR link is included then a Status should also be included.

Everything else LGTM. Thanks for the changes!

PR #53, opened 2021-08-02
Updated the status of some tests

mohasarc:
I've opened this PR https://github.com/vojtechhabarta/typescript-generator/pull/688 just over a month ago; at the time I was thinking it might get accepted or rejected then I would update the DB, but it didn't yet, and I forgot about it. Hence, now I'm updating the status of the tests to `Opened`.


winglam:
LGTM. Thanks for the changes!

PR #97, opened 2021-10-11
Opened PRs for `TestDeleteByPrimaryKey` in `abel533/Mapper`

Talank:
Starting from the patch suggested by the iFixFlakies tool, I did some debugging to find what is causing the tests to fail. And it turned out that there is no data at the database when the victim test is executed after polluters in a shared JVM. So, I added a `Before` method to reset the database and it solves all the flaky tests in the test class `tk.mybatis.mapper.test.country.TestDeleteByPrimaryKey`


darko-marinov:
Thanks for this PR. Please change the title to something like "Opened PRs for `abel533/Mapper`" or "Opened PRs for `TestDeleteByPrimaryKey` in `abel533/Mapper`", and I can accept this PR. We use `WIP` just for students to mark work in progress so they don't duplicate the work on fixing the same test(s). 

As for the actual PR (https://github.com/abel533/Mapper/pull/838), I'll send an email.

PR #154, opened 2021-11-04
NonDex does not report flaky on <org.zalando.riptide.failsafe.FailsafePluginRetriesTest.shouldAllowNestedCalls>

SpiceOwlf:
I ran NonDex on current SHA(fd15f72892bef56d27d0d8715d48c6c4d5a6655c) with -DnondexRuns=10, and it passed it. Then I ran NonDex on the reported-flaky SHA(8277e11fc069d8e24df0d233ef2577cc75659b75) also with with -DnondexRuns=10, and it still passed it.  I recored the output from SHA(8277e11fc069d8e24df0d233ef2577cc75659b75) as nondex_01.log in my VM right not. Please let me know what we can do to correct this.


SpiceOwlf:
I deleted the inaccurate line. The test is not ID flaky based on NonDex report.

PR #157, opened 2021-11-04
Update on riptide, shouldUseCredentialsDirectory

SpiceOwlf:
Developer has changed test location in the latest SHA(fd15f72892bef56d27d0d8715d48c6c4d5a6655c). Now it is under:

- riptide-spring-boot-autoconfigure

- org.zalando.riptide.autoconfigure.AccessTokensDefaultCredentialsDirectoryTest.shouldUseCredentialsDirectory

Furthermore, I tried to run this test with -Dtest=..., but MVN is complaining that it can not find it. Would like to discuss more over Campuswire/email


SpiceOwlf:
Made an update on [org.zalando.riptide.autoconfigure.AccessTokensCredentialsDirectoryTest#shouldUseCredentialsDirectory]. In the latest SHA, it has different path and file name, and it also passed NonDex test with -DnondexRuns=10.

darko-marinov:
Squashed several commits

PR #210, opened 2021-11-13
Checker improvement

SpiceOwlf:
Added two lines to tell users where the order checker fail, and where they can find related commands to fix them.


winglam:
Thanks for improving the checker!

PR #279, opened 2021-11-24
update pr status

shunfan-shao:
Will maintain opened (but not accepted) pull request in this branch.


shunfan-shao:
```
org.apache.zeppelin.iotdb.IoTDBInterpreterTest.testShowAllTTL
org.apache.zeppelin.iotdb.IoTDBInterpreterTest.testShowTimeseries
org.apache.zeppelin.iotdb.IoTDBInterpreterTest.testShowStorageGroup
```

The rest of the three tests in the same module have similar issues. I decided to not fix them because I am not do not understand some lower level concepts of SQL, eg: does "show tables" guarantee a consistent order? According to the test failures, it does not. But I cannot find any official references. 

I tried "debug" with nondex tool, it did not generate a trace file. And

PR #295, opened 2021-11-26
The PR checker should check DeveloperFixed Notes

pthariensflame:
Previously this wasn't happening, as visible in #294.


pthariensflame:
@winglam

format_checker/pr_checker.py
```diff
@@ -91,7 +91,7 @@ def check_status_consistency(filename, row, i, log):
         else:
             check_pr_link(filename, row, i, log)
 
-    if row["Status"] in ["InspiredAFix", "Skipped", "MovedOrRenamed"]:
+    if row["Status"] in ["InspiredAFix", "Skipped", "MovedOrRenamed", "DeveloperFixed"]:
```
winglam, line 94: I assume that the desire to require Notes for DeveloperFixed status is to have people provide the information for when a flaky test was fixed.

Although, it is certainly nice if people would provide Notes detailing when the developer fixed a particular flaky test, I don't envision that being easy for most to provide. E.g., for someone that runs some flaky test detection tool on the latest version of the project and finds that a particular flaky test is no longer flaky, we do not want to limit the contribution of such information only if that person provides Notes to describe when the test is fixed by the developer. 

Unless there are other reasons to add this requirement, I suggest removing this change from this PR.

winglam (commented):
Thanks for your changes!

PR #300, opened 2021-11-26
Contribute more ID tests

shunfan-shao:


shunfan-shao:
Test Name| 933178| 974622| 1016066  
---|---|---|---  
org.skyscreamer.jsonassert.comparator.CustomComparatorTest#testFullArrayComparison|✖| ✔| ✖


The other tests fail not only in nondex but also normal environment. It might be some misconfiguration.

PR #334, opened 2021-11-30
Draft: Add pr-check for forked project

shunfan-shao:
Add some checks for forked project. Please let me know if there is any issue

### DIFF
--- a/format_checker/common_checks.py
+++ b/format_checker/common_checks.py
@@ -38,8 +38,8 @@ def check_repo_sanity(checked_projects, filename, row, i, log):
     try:
         resp = requests.get(url).json()
         # Determine if it is a forked project
-        if "fork" in resp and resp["fork"]:
-            log_std_error(filename, log, i, row, "Project URL")
+        if resp.get("fork"):
+            log_esp_error(filename, log, f"{author}/{repo} is a forked repo")
     except requests.exceptions.RequestException as e:
         # handle(e)
         pass


shunfan-shao:
Currently GitHub API is fully public. There is an hour limit for specific IP address. 

Here are some information on GitHub API rate limit: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting

shunfan-shao:
Add some more details per discussed. The pipeline should fail when I explicitly committed a change with project being a forked one: https://github.com/TestingResearchIllinois/idoft/runs/4378193026?check_suite_focus=true.

format_checker/common_checks.py
```diff
@@ -14,7 +15,7 @@
 
 # Contains regexes for columns that are commmon to pr-data and tic-fic-data
 common_data = {
-    "Project URL": re.compile(r"(https:\/\/github.com)(\/(\w|\.|-)+){2}"),
+    "Project URL": re.compile(r"(https:\/\/github.com\/([\w|\.|-]+)\/([\w|\.|-]+))"),
```
winglam, line 18: Was the unchanged line giving you some issues?

format_checker/common_checks.py
```diff
@@ -23,6 +24,27 @@
...
+
+    url = "https://api.github.com/repos/{}/{}".format(author, repo)
+    try:
+        resp = requests.get(url).json()
+        # Determine if it is a forked project
+        if resp.get("fork"):
+            log_esp_error(filename, log, f"{author}/{repo} is a forked repo")
+    except requests.exceptions.RequestException as e:
+        # handle(e)
+        pass
```
winglam, line 45: Instead of simply passing here, it would be good to warn that the check for fork failed along with outputting the exception

format_checker/common_checks.py
```diff
@@ -123,6 +146,9 @@ def run_checks(file, data_dict, log, commit_range, checks):
                         if check_rule.__name__ == check_row_length.__name__:
                             check_rule(len(header), *params)
                             continue
+                        if check_rule.__name__ == check_repo_sanity.__name__:
```
winglam, line 149: Instead of running this check_rule for every changed line in the CSV file, we could just run this check when a new Project URL is added that does not already exist. i.e., get (1) the Project URLs that are already in the unchanged file and (2) the Project URLs in the new changes, and only run this check for Project URLs that are in (2) but not in (1).

This improvement need not be done immediately but would help avoid the Github limit.

winglam (commented):
Thanks for the changes! Minor suggestions to improve.

PR #374, opened 2021-12-03
Update pr status

shunfan-shao:


shunfan-shao:
In current pr-data.xml, there are two duplicate tests: https://github.com/TestingResearchIllinois/idoft/blob/main/pr-data.csv#L2791. 

The latest commits removed one of them...

darko-marinov:
Squashed three commits

PR #383, opened 2021-12-04
Add RepoArchived check for PR

LeoYimingLi:
(a) Changes overview:
- format_checker/common_checks.py 
  - check the archived repo in proposed PR without ```RepoArchived``` status
- format_checker/pr_checker.py
  - check the non-archived repo in proposed PR but with ```RepoArchived``` status
- format_checker/requirements.txt
  - add one dependency
- format_checker/utils.py 
  - add a log function for repo-archived error
 
(b) Some test results showing the changes work by running ```python3 format_checker/main.py```
1. check the archived repo in proposed PR without ```RepoArchived``` status
image

2. check the non-archived repo in proposed PR but with ```RepoArchived``` status
image

### DIFF
--- a/format_checker/common_checks.py
+++ b/format_checker/common_checks.py
@@ -3,12 +3,14 @@
 import re
 import csv
 import subprocess
+import requests
 from utils import (
     get_committed_lines,
     get_uncommitted_lines,
     log_info,
     log_std_error,
     log_esp_error,
+    log_archived_error,
 )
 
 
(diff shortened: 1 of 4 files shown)


format_checker/common_checks.py
```diff
@@ -23,6 +25,18 @@
...
+def check_repo_archived(checked_projects, filename, row, i, log):
+    project_url = row["Project URL"]
+    try:
+        resp = requests.get(project_url)
+        # Determine if it is an archived project
+        if "This repository has been archived by the owner. It is now read-only." in resp.text:
+            log_archived_error(filename, log, i, row, "Project URL")
+    except requests.exceptions.RequestException as e:
+        # handle(e)
+        pass
```
winglam, line 37: Instead of simply passing here, it would be good to warn that the check for repo archived failed along with outputting the exception

winglam (commented):
As this PR is quite similar to https://github.com/TestingResearchIllinois/idoft/pull/334 I would prefer merging that one first and then for some caching of `requests.get(project_url)` to be used so that at most we only make one request for each unique Project URL instead of making one for fork checking, one for repo archive, etc.

format_checker/common_checks.py
LeoYimingLi: get it, I change ```pass``` to ```log_esp_error(filename, log, "The check for repo-archived failed due to ERROR:" + str(e))```

yes the PR is similar to 334, maybe the two funcitions can merge into one ?

PR #401, opened 2021-12-06
pr-status update on karate

shunfan-shao:


shunfan-shao:
- delete one duplicate test
- update status for others

PR #427, opened 2021-12-08
Add scripts to automatically detect deleted tests

FanYuliang:
* `find_all_deleted.py` can output a json file (`new_test_file_deleted.json`) that contains the fully-qualified Test name and the first SHA that deletes that test file, which helpes developer find the deleted tests. 
* Each entry in the json contains three elements: `URL, packageName.ClassName.methodName, SHA`
* Further automated/ manual checks are required to determine if the test belongs to `Deleted` or `MovedOrRenamed`; useful git command: `git show <SHA> | grep -C <methodName/method content>`


auto-check-deleted/find_all_deleted.py
```diff
@@ -0,0 +1,71 @@
...
+	print("clone done!!!")
+	for full_testname, status, note in zip(testname_list, status_list, note_list):
+		testfile_name, test_method_name = full_testname.split(".")[-2], full_testname.split(".")[-1]
+		
+		if pd.isna(status) and pd.isna(note):
+			
+			if "[" in test_method_name or "\"" in test_method_name:
+				print("{}, {} is a PUT; skip".format(url, full_testname))
+				continue
+			if testfile_name[-4:] != "Test":
```
Ajitesh27, line 41: Many tests in pr-data.csv have "Test" in the beginning instead of the end.

PR #461, opened 2021-12-11
Add Python NIO tests from ICSE'22

Anjiang-Wei:
This PR aims to add 138 Python NIO tests from our ICSE'22 paper: `Preempting Flaky Tests via Non-Idempotent-Outcome Tests`. All the PR statuses are the latest (checked on Dec. 10).
Sorting script:
`echo "$(head -n1 py-data.csv && tail +2 py-data.csv | LC_ALL=C sort -k1,1 -k3,3 -t, -f)" > py-data.csv`
(first alphabetical order for Project URL, then Pytest Test Name)

``
The following script is for validation:
```
tail +2  py-data.csv | wc -l # header shouldn't be counted
for category in Accepted InspiredAFix Rejected Opened; do
    num=$(grep $category py-data.csv | wc -l)
    echo $category:$num
done
```
The result:
```
138
Accepted:34
InspiredAFix:1
Rejected:1
Opened:46
```
Submitted: 34+1+1+46=82
The numbers are now consistent with the paper: total=138, submitted=82, accepted=35, rejected=1
Sidenote: we count the `InspiredAFix` as `Accepted` in the paper, which will be discussed in the final version of the paper(\Fix added).


py-data.csv
```diff
@@ -0,0 +1,139 @@
+Project URL,SHA Detected,Pytest Test Name (PathToFile::TestClass::TestMethod or PathToFile::TestMethod),Category,Status,PR Link,Notes
```
winglam, line 1: Should it be `pytest` instead of `Pytest`?

Is the idea of the `Pytest Test Name` to give a name that can be given directly to `pytest` to run the test later? (i.e., `pytest [Pytest Test Name]` should work for all test names in this file)

How specific to `pytest` are the `Pytest Test Name`? i.e., do other Python frameworks express test names in a completely different format? Should this column be called `Test Name` instead?

Anjiang-Wei: I think this is specific to `pytest`: https://docs.pytest.org/en/latest/how-to/usage.html, (`pytest test_mod.py::TestClass::test_method`)
In https://docs.python.org/3/library/unittest.html (Python unittest), tests are invoked in a different way, e.g., `python -m unittest test_module.TestClass.test_method`.
In https://stackoverflow.com/questions/25005277/how-to-run-specific-test-cases-from-a-test-suite-using-robot-framework (Robot, another popular Python testing framework), tests are invoked in yet another way, e.g., `robot -t testcase1 mytestsuite.robot`.

> Is the idea of the Pytest Test Name to give a name that can be given directly to pytest to run the test later? (i.e., pytest [Pytest Test Name] should work for all test names in this file)

Yes. It can be directly passed to pytest.

> Should it be pytest instead of Pytest

I use `Pytest` because the all initials in the header are capitalized. But I am happy change if you think `pytest` is the better choice. What's your final judgement call?

winglam: Thanks for the clarifications.

It's fine that the test name is specific to pytest and to leave pytest capitalized in the header.

The changes look fine to me now. We can address other concerns later if there are any.

PR #466, opened 2021-12-23
Remove redundant lines for NIO tests

Anjiang-Wei:
The script for checking redundant lines for NIO tests are `grep NIO pr-data.csv | cut -d, -f4 | sort | uniq -c | sort -rn | head -9`.
I resolve redundant lines by either removing the duplicate line or the line with less info, or merging two lines into one.

After resolving, the left duplicate lines have two separate PRs opened, which should not be merged.


winglam:
LGTM. Thanks for the fix!

PR #507, opened 2022-03-24
1 new accepted entry in wildfly/jboss-ejb-client

kaiyaok2:
As title


kaiyaok2:
The auto check says my PR link is invalid: https://github.com/wildfly/jboss-ejb-client/pull/534
The link works find on my end though.

PR #523, opened 2022-04-08
Fix a flaky test and inspire a fix for another

chessvivek:
PR - https://github.com/spring-projects/spring-boot/pull/30514


chessvivek:
Converted to draft because it seems like developers made a minor change to my fix that makes the tests flaky again.

chessvivek:
Marking ready for review again because the developer confirmed that the particular minor change does not lead to any non-determinism.

PR #542, opened 2022-04-23
Adding link to another PR for InternalThreadLocalTest

august782:
Noting accepted PR for `org.apache.dubbo.common.threadlocal.InternalThreadLocalTest.testSize`. Note that this test was marked as NIO and had an earlier PR fixing just the `testSize` test, but it is actually order-dependent w.r.t. other tests in the same test class (so it fails when run after some other tests). The new PR essentially puts the reset in a setup method, which was accepted. For now, I use some `|` to separate the two PRs, but unsure what should be the style for indicating the two PRs per test.


winglam:
Did you find `org.apache.dubbo.common.threadlocal.InternalThreadLocalTest.testSize` to be OD in `737f7a7ea67832d7f17517326fb2491d0a086dd7`?

If not, the procedure here should probably be to just add a new line for this test with the SHA that the test was actually found to be OD.

If the test is already OD in `737f7a7ea67832d7f17517326fb2491d0a086dd7`, then the `|` separator looks fine to me

august782:
Running iDFlakies on my end was able to detect this flaky test on that commit. I've tried changing it to put the other PR link in the Notes column. Not sure if it is necessary to update the type for the test, i.e., should it be modified to indicate it is also OD-Vic?

darko-marinov:
Thanks for the changes. The test was already marked as both OD-Vic and NIO.

PR #641, opened 2022-11-17
Update Moved and New tests

BoyuLi4:
The old model of these tests `guava-test` should be `guava-tests`
Use`git log --full-history -- guava-test` to check, there is no history of `guava-test`
So, count these tests as "Renamed"
(Used the old SHA in the Note column, since there is no "change" commit)

Add new tests with the last commit of this model.
```
cd guava
mvn install -pl guava/tests -am -DskipTests
cd guava/tests
mvn nondex:nondex
```
Found file `failures`

```
com.google.common.collect.SetsFilterHashSetTest#testForEach
com.google.common.collect.MultimapsTest#testUnmodifiableHashMultimap

```
Rechecked by:
`mvn -pl guava/tests edu.illinois:nondex-maven-plugin:2.1:nondex -Dtest=com.google.common.collect.SetsFilterHashSetTest#testForEach`

These two are new flaky tests.
Other `Renamed `Build Success.


pr-data.csv
```diff
@@ -2557,16 +2557,18 @@ https://github.com/fromage/redpipe,0aff891d6befdf0dcc2bdbfda22262cdf82ac66f,redp
...
-https://github.com/google/guava,a197d99b96d69ea75a2c4a15ecd8a730dd3270d6,guava-test,com.google.common.collect.MapsTest.testToStringImplWithNullValues,ID,,,
-https://github.com/google/guava,a197d99b96d69ea75a2c4a15ecd8a730dd3270d6,guava-test,com.google.common.collect.MultimapsTest.testUnmodifiableArrayListMultimap,ID,,,
-https://github.com/google/guava,a197d99b96d69ea75a2c4a15ecd8a730dd3270d6,guava-test,com.google.common.collect.MultimapsTest.testUnmodifiableHashMultimap,ID,,,
-https://github.com/google/guava,a197d99b96d69ea75a2c4a15ecd8a730dd3270d6,guava-test,com.google.common.collect.MultimapsTest.testUnmodifiableSynchronizedArrayListMultimap,ID,,,
-https://github.com/google/guava,a197d99b96d69ea75a2c4a15ecd8a730dd3270d6,guava-test,com.google.common.collect.MultimapsTest.testUnmodifiableSynchronizedHashMultimap,ID,,,
+https://github.com/google/guava,a197d99b96d69ea75a2c4a15ecd8a730dd3270d6,guava-test,com.google.common.collect.FilteredCollectionsTest$SetsFilterHashSetTest.testForEach,ID,MovedOrRenamed,,https://github.com/google/guava/commit/a197d99b96d69ea75a2c4a15ecd8a730dd3270d6
```
darko-marinov, line 2560: Did you want to change all these to `guava-tests`?

darko-marinov (changes requested):
Please change all `guava-test` to `guava-tests`, and then you can also add `MovedOrRenamed`.

pr-data.csv
BoyuLi4: So sorry...how can I used the wrong column of the CSV file...

BoyuLi4: Hi,
I changed it.
But for tests below line 2572, I'm not sure how to test it since the test name seems not normal.
`testToStringMatchesAsMap[ArrayListMultimap [collection size: several] reserialized [collection size: several]]`

So I'm not sure what state should be imposed on them. Or should I just change them into `tests` first?

darko-marinov: Change to tests. Those are so-called "parameterized unit tests". See in the source how the test is written.

BoyuLi4: Thanks for the notice. 
I might check on that later.
These tests are actually from another place `guava-testlib`.....not just `test` to `tests`
And some of them are in "src" not "test"

PR #644, opened 2022-11-17
add progress on flaky tests

anantdahiya8:
Adding progress up till progress report 4


pr-data.csv
```diff
@@ -3392,6 +3402,13 @@ https://github.com/openpojo/openpojo,9badbcc4593e797accfed5e51bec9f2b843f0f67,.,
...
 https://github.com/openpojo/openpojo,9badbcc4593e797accfed5e51bec9f2b843f0f67,.,com.openpojo.validation.test.impl.SetterTesterTest.shouldSuccessfullyValidateEvenIfLoggingFails,OD;UD,Opened,https://github.com/OpenPojo/openpojo/pull/136,
 https://github.com/openpojo/openpojo,9badbcc4593e797accfed5e51bec9f2b843f0f67,.,com.openpojo.validation.utils.ValidationHelperTest.shouldReportMissingASMProperly,OD-Vic,Opened,https://github.com/OpenPojo/openpojo/pull/136,
+https://github.com/OpenRefine/OpenRefine,3bff27b868a6a1b695f3855ae0f49e269c5ef7cd,main,com.google.refine.browsing.facets.ListFacetTests.serializeListFacet,ID,,,
+https://github.com/OpenRefine/OpenRefine,3bff27b868a6a1b695f3855ae0f49e269c5ef7cd,main,com.google.refine.browsing.facets.ListFacetTests.testSelectedEmptyChoice,ID,,,
+https://github.com/OpenRefine/OpenRefine,3bff27b868a6a1b695f3855ae0f49e269c5ef7cd,main,com.google.refine.clustering.binning.BinningClustererTests.testSerializeBinningClusterer,ID,Opened,https://github.com/OpenRefine/OpenRefine/pull/5426,
```
darko-marinov, line 3407: You should follow up on this; it's been a week since the developer provided feedback.

pr-data.csv
```diff
@@ -2284,8 +2284,18 @@ https://github.com/doanduyhai/Achilles,e3099bdce342910951c4862c78751fd81ed4552e,
...
+https://github.com/dromara/hutool,ec14f8384ad384fb6e054df722fe089d64af0c41,hutool-core,cn.hutool.core.annotation.AnnotationUtilTest.scanMetaAnnotationTest,ID,,,
+https://github.com/dromara/hutool,ec14f8384ad384fb6e054df722fe089d64af0c41,hutool-core,cn.hutool.core.annotation.GenericSynthesizedAggregateAnnotationTest.baseSynthesisAnnotationWorkTest,ID,,,
 https://github.com/dromara/hutool,12969f248329f81f76eb5bece8c718ac1c63f7b1,hutool-core,cn.hutool.core.text.StrJoinerTest.joinMultiArrayTest,OD,Accepted,https://github.com/dromara/hutool/pull/1883,
 https://github.com/dromara/hutool,12969f248329f81f76eb5bece8c718ac1c63f7b1,hutool-core,cn.hutool.core.util.JAXBUtilTest.beanToXmlTest,OD,Accepted,https://github.com/dromara/hutool/pull/1851,
+https://github.com/dromara/hutool,ec14f8384ad384fb6e054df722fe089d64af0c41,hutool-core,cn.hutool.core.util.ReflectUtilTest.getMethodsFromInterfaceTest,ID,,,
+https://github.com/dromara/hutool,ec14f8384ad384fb6e054df722fe089d64af0c41,hutool-json,cn.hutool.json.Issues1881Test.parseTest,ID,DeveloperWontFix,,
```
darko-marinov, line 2297: Why is this `DeveloperWontFix`?

darko-marinov (changes requested):
Please clarify some entries, or split the PR into two: one that can be accepted right away and another for potentially longer discussion.

pr-data.csv
anantdahiya8: https://github.com/dromara/hutool/pull/2713

pr-data.csv
anantdahiya8: Developer asked not to make change in the main code as it may cause some performance issue and try to solve the flaky test by making changes in the test code. But it is not that straight forward to fix in the test code since the code which has the assert is part of a generic test code which is used by multiple tests, some which which contain bigger jsons. I am trying to find a fix which does not end up breaking other testcases

pr-data.csv
xieyt2000: The final comment in English:
Fine, leave it be until there're problems. The test has no serious problem.
BTW, I think you should also include the PR Link in the two rows

xieyt2000 (changes requested):

PR #635, opened 2022-11-14
Added Git Bisect Script

akshathsk:
This PR contains the script used to find the commit which fixed the flaky test https://github.com/TestingResearchIllinois/idoft/pull/633

References -
1. https://git-scm.com/docs/git-bisect/
2. https://docs.google.com/presentation/d/1m5asEw506nfqthuTsQfF5_3BBLvtm2edN3dnFLMqtlY/edit?usp=sharing

PR #647, opened 2022-11-18
Some Updates to pr-data.csv

Kerr0220:
These are tests that have been found to be non-flaky or has been accepted/rejected.
Please have look on them!


pr-data.csv
```diff
@@ -809,10 +809,10 @@ https://github.com/apache/hadoop,cc2babc1f75c93bf89a8f10da525f944c15d02ea,hadoop
...
 https://github.com/apache/hadoop,cc2babc1f75c93bf89a8f10da525f944c15d02ea,hadoop-hdfs-project/hadoop-hdfs-nfs,org.apache.hadoop.hdfs.nfs.nfs3.TestWrites.testWriteStableHow,OD-Vic,,,
 https://github.com/apache/hadoop,cc2babc1f75c93bf89a8f10da525f944c15d02ea,hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient,org.apache.hadoop.hdfs.TestNNBench.testNNBenchCreateReadAndDelete,ID,,,
-https://github.com/apache/hadoop,14cd969b6ea1898e9db6eeb9ea5292ec4558a706,hadoop-hdfs-project/hadoop-hdfs-client,org.apache.hadoop.hdfs.web.oauth2.TestClientCredentialTimeBasedTokenRefresher.refreshUrlIsCorrect,ID,,,
-https://github.com/apache/hadoop,14cd969b6ea1898e9db6eeb9ea5292ec4558a706,hadoop-hdfs-project/hadoop-hdfs-client,org.apache.hadoop.hdfs.web.oauth2.TestRefreshTokenTimeBasedTokenRefresher.refreshUrlIsCorrect,ID,,,
+https://github.com/apache/hadoop,14cd969b6ea1898e9db6eeb9ea5292ec4558a706,hadoop-hdfs-project/hadoop-hdfs-client,org.apache.hadoop.hdfs.web.oauth2.TestClientCredentialTimeBasedTokenRefresher.refreshUrlIsCorrect,ID,DeveloperFixed,,
```
xieyt2000, line 812: In which commit did the developer fix the test? You can have a look at https://github.com/TestingResearchIllinois/idoft/pull/635

pr-data.csv
```diff
@@ -809,10 +809,10 @@ https://github.com/apache/hadoop,cc2babc1f75c93bf89a8f10da525f944c15d02ea,hadoop
...
+https://github.com/apache/hadoop,14cd969b6ea1898e9db6eeb9ea5292ec4558a706,hadoop-hdfs-project/hadoop-hdfs-client,org.apache.hadoop.hdfs.web.oauth2.TestRefreshTokenTimeBasedTokenRefresher.refreshUrlIsCorrect,ID,DeveloperFixed,,
 https://github.com/apache/hadoop,a585a73c3e02ac62350c136643a5e7f6095a3dbb,hadoop-hdfs-project/hadoop-hdfs-client,org.apache.hadoop.hdfs.web.TestURLConnectionFactory.testSSLFactoryCleanup,OD,,,
-https://github.com/apache/hadoop,14cd969b6ea1898e9db6eeb9ea5292ec4558a706,hadoop-common-project/hadoop-common,org.apache.hadoop.ipc.TestDecayRpcScheduler.testPriority,ID,,,
+https://github.com/apache/hadoop,14cd969b6ea1898e9db6eeb9ea5292ec4558a706,hadoop-common-project/hadoop-common,org.apache.hadoop.ipc.TestDecayRpcScheduler.testPriority,ID,Rejected,https://github.com/apache/hadoop/pull/5085,
```
xieyt2000, line 815: Why is this rejected? The PR is still open and it needs some followup.

xieyt2000 (changes requested):

pr-data.csv
darko-marinov: Indeed, there was no follow up for 3 weeks.

pr-data.csv
Kerr0220: Oh, these tests are not flaky now. I've checked possible status of tests in idoft and suppose that these cases are suitable for `DeveloperFixed`. Should I change them to anther status?

pr-data.csv
Kerr0220: And I'm not quite sure about the answer of the question raised by the developer that in which case/scenario the test will fail.

pr-data.csv
xieyt2000: Yes, they're `DeveloperFixed` if they're still there but not flaky now. I'm saying it's better to find which commit did the developer fix the test

pr-data.csv
darko-marinov: What is the root cause of the failure? Does it fail because of (1) the iteration order of `HashMap`, (2) `getDeclaredFields`, or (3) something third?

Kerr0220: It's caused by the process of converting an object to a string. The attributes of that object were converted into a non-deterministic order.

Kerr0220:
Another PR will be made instead of this one.

Kerr0220:
Here's the new [PR-682](https://github.com/TestingResearchIllinois/idoft/pull/682)

PR #659, opened 2022-11-25
Update Test data

bmk15897:
I manually checked the File history to find the commits where the methods were renamed. 
But the body of the tests was different than the earlier version and I found that the newer version of the tests is not flaky. So I am not adding new rows in the IDOFT PR Data for the newer version.


xieyt2000 (approved):
I agree this should be removed.

darko-marinov:
If the tests are not flaky, put `DeveloperFixed` and find where they were fixed.

bmk15897:
Apologies, I mistook these for ID tests. As they are UD, I do not know whether they are fixed. Shall I add new rows for these tests?

darko-marinov:
Hard to tell when/if the `UD` tests were fixed, because we had a hard time reproducing them. If they were `MovedOrRenamed` or `Deleted`, just document the commit where it was done. It'd be hard to label as `DeveloperFixed`.

bmk15897:
I have already documented the commit where the tests were deleted. Acc. to IDOFT README, if the test body is different for the renamed tests, the original test should be marked as Deleted and new row should be added with the renamed test. So I should be adding new rows in the pr-data.csv for the renamed tests as it is hard to check the flakiness of UD tests?

darko-marinov:
You can check the `OD-Brit` test `org.apache.dubbo.config.AbstractInterfaceConfigTest.checkStubAndMock6`. The others were `UD` and can be marked just as `Deleted`.

bmk15897:
I ran the test org.apache.dubbo.config.AbstractInterfaceConfigTest.checkStubAndMock6 on its current version and it passed.

darko-marinov:
Was it `OD-Brit` in the reported version?

bmk15897:
No, it wasn't. I ran that one too and it passed.

PR #681, opened 2022-12-02
add tab checker

ZhewenFu:
Add a checker in `pr_checker.py` to check if there is a tab in the fields.

### DIFF
--- a/format_checker/pr_checker.py
+++ b/format_checker/pr_checker.py
@@ -130,7 +130,7 @@ def check_status_consistency(filename, row, i, log):
 
     if row["Status"] == "" and row["PR Link"] != "":
         check_pr_link(filename, row, i, log)
-        log_std_error(filename, log, i, row, "Status should not be empty when a PR link is provided.")        
+        log_std_error(filename, log, i, row, "Status", "Status should not be empty when a PR link is provided.")        
 
 def check_notes(filename, row, i, log):
     """Checks validity of Notes."""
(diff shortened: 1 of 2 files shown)


format_checker/pr_checker.py
```diff
@@ -148,6 +148,13 @@ def check_pr_link(filename, row, i, log):
     ):
         log_std_error(filename, log, i, row, "PR Link")
 
+def check_tab(filename, row, i, log):
+    """Checks that there is no tab in the row."""
+    
+    for key,value in row.items():
+        if '\t' in value:
+            log_std_error(filename, log, i, row, key)
```
darko-marinov, line 156: Can you provide some descriptive message for why this row fails? Did you test this on some examples?

ZhewenFu: There is no error message field in `log_std_error`. Should I use `log_warning` instead? Also, I have noticed that in the function `check_status_consistency` line 133, it used `Status should not be empty when a PR link is provided.` as error message which is not supported in `log_std_error`. 

I have tested this on some examples locally.

darko-marinov: I don't know the details of this script, but if you think `log_warning` is better, please use that. Also, if you want to improve/correct line 133, please do so.

@xieyt2000 can you please check that line 133?

xieyt2000: I think the line 133 has a bug since it should give `key` in the last parameter.
@ZhewenFu Can you make the `log_std_error` support error message? I don't think we should use `log_warning` here

xieyt2000 (changes requested):

xieyt2000 (approved):

darko-marinov:
Accepting per Yuntong's approval.

PR #685, opened 2022-12-04
New ID test detected from fastjson2

tt0suzy:
I only realized these tests are all carried from the original fastjson repo after detecting them, so you might not want to add them. But they are now in a new repo so I'm opening the pr anyways. 

Log file from NonDex:
[fastjson.log](https://github.com/TestingResearchIllinois/idoft/files/10148902/fastjson.log)


darko-marinov:
It's not interesting to add duplicates, but `fastjson` did have unusual history (e.g., now two repos `fastjson` and `fastjson2`, none being forked; previously some rewrite of commits so that old commits disappeared). Can you check if *all* tests from `fastjson` also appear in `fastjson2`?

tt0suzy:
I don't know how to write a script to check it but I just randomly picked 2 tests `testIssue3493` `SerializeWriterJavaSqlDateTestfrom` from `fastjson` and they are not in `fastjson2`, so I think not all tests from `fastjon` are carried to `fastjson2`.

PR #688, opened 2022-12-07
Update test status

sopan98:


pr-data.csv
```diff
@@ -2029,99 +2029,98 @@ https://github.com/apache/ignite-3,a1aa7fd4e2398c72827777ec5417d67915ff4da3,modu
...
 https://github.com/apache/ignite-3,a1aa7fd4e2398c72827777ec5417d67915ff4da3,modules/core,org.apache.ignite.internal.tostring.IgniteToStringBuilderSelfTest.testHierarchy,ID,,,
 https://github.com/apache/ignite-3,a1aa7fd4e2398c72827777ec5417d67915ff4da3,modules/core,org.apache.ignite.internal.tostring.IgniteToStringBuilderSelfTest.testToString,ID,,,
 https://github.com/apache/ignite-3,a1aa7fd4e2398c72827777ec5417d67915ff4da3,modules/core,org.apache.ignite.internal.tostring.IgniteToStringBuilderSelfTest.testToStringCheckConcurrentModificationExceptionFromMap,ID,,,
 https://github.com/apache/ignite-3,a1aa7fd4e2398c72827777ec5417d67915ff4da3,modules/core,org.apache.ignite.internal.tostring.IgniteToStringBuilderSelfTest.testToStringSizeLimits,ID,,,
 https://github.com/apache/ignite-3,a1aa7fd4e2398c72827777ec5417d67915ff4da3,modules/core,org.apache.ignite.internal.tostring.IgniteToStringBuilderSelfTest.testToStringWithAdditions,ID,,,
-https://github.com/apache/ignite-3,a1aa7fd4e2398c72827777ec5417d67915ff4da3,modules/core,org.apache.ignite.internal.util.worker.IgniteWorkerTest.testUpdateHeartbeat,ID,,,
```
sopan98, line 2121: I ran nondex on this SHA for for this test and all these tests were passing.

darko-marinov: Did you run the entire test class? Did you run the entire module? Did you run the entire project? Did you run for more than three seeds?

I don't think you need to add these late changes to this same PR. Let's first finalize and accept what you had before, and then you can open another PR later.

sopan98: I had run the entire module first with 10 seeds. I did not see these tests in the output. So I ran each test with 100 seeds. They still passed.

darko-marinov: Thanks for confirming. Let's accept this PR so we can move on.

PR #708, opened 2022-12-12
Adding new category for Portability tests

bmk15897:
Portability tests are those which fail or pass depending upon the OS that they are run on.

### DIFF
--- a/format_checker/pr_checker.py
+++ b/format_checker/pr_checker.py
@@ -33,6 +33,7 @@ pr_data = {
         "NDOI",
         "NDOI",
         "UD",
+        "Portability"
     ],
     "Status": [
         "",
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -124,6 +124,7 @@ NOD | Non-Deterministic tests
 NDOD | Non-Deterministic Order-Dependent tests that fail non-deterministically but with significantly different failure rates in different orders as defined in our [ISSRE’20 work](http://mir.cs.illinois.edu/winglam/publications/2020/LamETAL20ISSRE.pdf)
 NDOI | Non-Deterministic Order-Independent tests that fail non-deterministically but similar failure rates in all orders as defined in our [ISSRE’20 work](http://mir.cs.illinois.edu/winglam/publications/2020/LamETAL20ISSRE.pdf)
 UD | Unknown Dependency tests that pass and fail in a test suite or in isolation
+Portability | Tests that pass and fail depending on the operating system
```
darko-marinov, line 127: How about calling them `OSD` for `Operating System Dependent`? Portability may be too broad as it can depend on many aspects of a platform not just OS.

format_checker/pr_checker.py
```diff
@@ -33,6 +33,7 @@
         "NDOI",
         "NDOI",
         "UD",
+        "Portability"
```
darko-marinov, line 36: `OSD` may be a better name; see below.

darko-marinov (changes requested):
`OSD` may be a better (more precise) name than `Portability`.

readme.md
bmk15897: Okay, will change that.

xieyt2000 (approved):

PR #720, opened 2022-12-13
Make it compatible with Java 9+

Jinzhengxu:
make this `./runNondex.sh` can chose different version of the Nondex.

### DIFF
--- a/auto-run-nondex/runNondex.sh
+++ b/auto-run-nondex/runNondex.sh
@@ -1,21 +1,14 @@
 #!/bin/bash
 DIR="${PWD}"
-echo $JAVA_Home
-version="1.1.2"
-read -p "Enter the nondex version you want run:(press 1 for nondex:1.1.2, press 2 for nondex:2.1.1)" name
-case $name in
-  1)
-    version="1.1.2"
-    echo "Chosen nondex:1.1.2"
-    ;;
-  2)
-    version="2.1.1"
-    echo "Chosen nondex:2.1.1"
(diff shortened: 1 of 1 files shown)


auto-run-nondex/runNondex.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 DIR="${PWD}"
+nondex_version="1.1.2"
+java_version=$(java -version 2>&1 | sed -n ';s/.* version "\(.*\)\.\(.*\)\..*".*/\1\2/p;')
+echo java_version "$java_version"
+if [[ "$java_version" -ge 90 ]]; then
+    echo version is more than 1.9
+    nondex_version="2.1.1"
```
darko-marinov, line 8: Version 2.1.1 supports Java 9+ (at least until 11) *and* Java 8. Why not always run with 2.1.1 instead of 1.1.2?

Jinzhengxu: I have completed the modification. As for why I don't always run with 2.1.1 instead of 1.1.2, I was thinking that some projects may require certain features that only 1.1.2 provides. If it weren't for these restrictions, we could certainly run with 2.1.1 directly.

darko-marinov: I'm not aware of any feature supported by 1.1.2 that's not supported by 2.1.1. Do you have some example? If yes, please share. If not, just change the version in one place below and remove these lines for checking the Java version.

Issue #731, opened 2023-03-09
https://github.com/apache/servicecomb-java-chassis, b79423c19f0bcf4e622674b5b74d9162a8af7a02, org.apache.servicecomb.swagger.generator.core.processor.annotation.SwaggerDefinitionProcessorTest.should_process_security_oauth2


Fring02:
Here's the flaky test detection in the repository https://github.com/apache/servicecomb-java-chassis on the SHA _b79423c19f0bcf4e622674b5b74d9162a8af7a02_. We've chosen that commit, because that commit is the first commit where the test method was introduced and got fail for flaky test. The previous commits before that simply didn't have that test method. Once it was introduced, Nondex detected flakiness.

For the ease of upload, the file has been changed to .txt format
[config.txt](https://github.com/TestingResearchIllinois/idoft/files/10926714/config.txt)
[failures.txt](https://github.com/TestingResearchIllinois/idoft/files/10926715/failures.txt)
[invocations.txt](https://github.com/TestingResearchIllinois/idoft/files/10926716/invocations.txt)
[org.apache.servicecomb.swagger.generator.core.processor.annotation.SwaggerDefinitionProcessorTest.txt](https://github.com/TestingResearchIllinois/idoft/files/10926717/org.apache.servicecomb.swagger.generator.core.processor.annotation.SwaggerDefinitionProcessorTest.txt)

Here's the usual execution of test, without Nondex. It succeeds, as it can be seen.

![text_execution_success](https://user-images.githubusercontent.com/63200885/223891413-197f2e64-dcb2-4417-b490-3f774b271e68.PNG)

We were working in a team of 3. Our members:
Sultanbek Khassenov (skhassen@gmu.edu) - @Fring02
Athena Parker (eparke2@gmu.edu) - @NegaNote
Mani Shah (mshah22@gmu.edu) - @mani9723

winglam:
Can you add some description for what each file you uploaded contains?

Fring02:
> Can you add some description for what each file you uploaded contains?

Done, thanks.

PR #732, opened 2023-03-11
Add issue for swagger test should_process_security_oauth2 to notes

NegaNote:
Hello.
Some preliminary work has been done to inspect the swagger test should_process_security_oauth2 (see #731), and I would like to add that issue to the CSV. I noticed that there isn't a proper Status for "inspected, but nothing else", so I kept the Status field blank -- is that acceptable?

Thank you!

Other team members:
@Mani9723
@Fring02


NegaNote:
Oh my god, I've noticed I put things in the wrong field -- I'm so sorry! Trying to fix with new PR.

PR #770, opened 2023-09-03
Update auto-update.yml, added check for gr-data.csv's format

Ellen99:
Updating auto-update.yml  check gr-data.csv file in addition to pr-data.csv


darko-marinov:
Is this `auto-update.yml` still running regularly? What does it even do? Sorry it's been a while since I looked into it.

Ellen99:
The workflow runs at 8:41 AM on the 4th day of every month (given in the schedule cron expression)

Ellen99:
ci.yml is responsible for running a format checker script (main.py) on push and pull requests

format_checker/common_checks.py
```diff
@@ -92,7 +92,7 @@ def check_sort(filename, log):
     diff = subprocess.check_output(command, shell=True).decode("utf-8")
     if diff != "":
         log_esp_error(filename, log, "The file is not properly ordered")
-        print("Refer to IDoFT readme for how pr-data.csv should be sorted: https://github.com/TestingResearchIllinois/idoft#to-contribute-a-newly-detected-flaky-test")
+        print("Refer to IDoFT readme for how pr-data.csv/gr-data.csv should be sorted: https://github.com/TestingResearchIllinois/idoft#to-contribute-a-newly-detected-flaky-test")
```
darko-marinov, line 95: Change `pr-data.csv/gr-data.csv` to `pr-data.csv or gr-data.csv` (to avoid confusion with a `directory_name/file_name`). More importantly, where should README mention what `gr-data.csv` is?

format_checker/gr_checker.py
```diff
@@ -0,0 +1,186 @@
+"""Implements rule checks for the gr-data.csv file."""
```
darko-marinov, line 1: I see more files now. Instead of having a copy of PR checker in GR checker, let's (1) make PR checker more general to take an input file name, and (2) run PR checker with both `pr-data.csv` and `gr-data.csv`.

darko-marinov (changes requested):
It's good to have more checks here. Let's just reduce the amount of copy-paste.

Ellen99:
I initially considered combining PR_checker and GR_checker into a single, more generalized script to handle both "pr-data.csv" and "gr-data.csv." 
However, I decided to maintain separate scripts for future flexibility. Separating the scripts allows us to modify and adapt the checking process for each file independently. If the format or criteria for one file needs a change in the future, we can make modifications to that specific script without impacting the other. Having separate scripts is also makes the code easier to understand and work with. 

But, if we are confident that we won't need to change the format of the files independently, I can generalize the PR_checker script.
Please let me know which solution you would prefer.

format_checker/common_checks.py
Ellen99: done

format_checker/gr_checker.py
Ellen99: https://github.com/TestingResearchIllinois/idoft/pull/770#issuecomment-1704336924

darko-marinov:
Let's keep one file for now and separate later if needed. I find it better than separating now thinking that we *may* need it later.

PR #776, opened 2023-09-23
add new status Claimed in format checker

zzjas:
:)


darko-marinov:
Okay to accept this as an undocumented feature to be removed at some point before the semester end.

PR #777, opened 2023-09-23
add forked projects entry for old projects

zzjas:
These projects didn't have forked-projects entries which breaks the format checker.


darko-marinov:
I'm accepting these projects, but if you check specifically these, maybe you can check *all* projects? :)

PR #788, opened 2023-10-04
Update pr-data.csv

KiruthikaJanakiraman:
1.com.networknt.client.Http2ClientTest.callApiWithHttp2ConnectionPoolAsync

This test passed with nondex in the latest commit as well as in the commit id given in Gitlab. I tried running the nondex command for 100 runs for which the test still passed. I also tried running nondex for the entire test class as well the entire test suite as suggested by professor Darko. The test still passes with nondex, hence I am raising this PR to remove this test from IDoFT.

2.com.networknt.client.Http2ClientTest.testSingleAsych

This test is getting skipped in the latest commit because the annotation `@Ignore` was added in commit https://github.com/networknt/light-4j/commit/6b3ce41ced026b08fa2db7f727a81f587252c1bf. Raising this PR to update the status of this test to `Deleted`.


darko-marinov:
For the second test, we need `Ignored` (or `Disabled`) instead of `Deleted`. You may (or not) split the PR into two, one that we can accept now, and another for later when the new status is added.

darko-marinov:
Also, for `RepoArchived`, if you want to check *all* the repos in IDoFT, that can be done (and there're some scripts around to help).

@zzjas you may also check if someone else wants to do this.

KiruthikaJanakiraman:
> For the second test, we need `Ignored` (or `Disabled`) instead of `Deleted`. You may (or not) split the PR into two, one that we can accept now, and another for later when the new status is added.

Sure, I have updated the PR by undoing the Deleted status for now.

KiruthikaJanakiraman:
> For the second test, we need `Ignored` (or `Disabled`) instead of `Deleted`. You may (or not) split the PR into two, one that we can accept now, and another for later when the new status is added.

I have updated the status if the Ignored test to `Skipped` according to the IDoFT Readme.

PR #806, opened 2023-10-10
change tests to opened and fixed

hofi1:
- org.wildfly.clustering.marshalling.spi.ExternalizerUtilTestCase.testUnmodifiableCollection


darko-marinov:
You need to open PR on the real repo on Zijie's fork.

PR #821, opened 2023-10-14
Add fixed flaky test for h2gis repo

219sansim:
The test passes with the latest SHA commit ID
I was not able to build the project with the commit ID `bc5ceab8e5ad82e8b6ec0b5c47898e45fa3d62b2` that the test was reported on but I tried with more recent commit IDs and the build succeeds and test becomes flaky for this commit ID - `7b7df4a597c154db88e03091d34e720de646f427`
Steps to reproduce-
1. `git clone https://github.com/orbisgis/h2gis.git`
2. `git checkout 7b7df4a597c154db88e03091d34e720de646f427`
3. `mvn clean install -DskipTests`
4. `mvn -pl h2gis-functions edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=org.h2gis.functions.io.dbf.DBFEngineTest#readDBFRussianWrongEncodingThenWriteThenRead`

**Note**-  The test passes when run for the first time and fails even if the test is run with mvn second time (even without NonDex)

**PR Fix** - https://github.com/orbisgis/h2gis/pull/1363


darko-marinov:
Please see email.

PR #846, opened 2023-10-20
Update pr data.csv

hofi1:
- eu.xenit.alfred.initializr.generator.docker.compose.DockerComposeYmlWriterDelegateTest.testService not reproducible
- org.apache.batchee.cli.MainTest.executions not reproducible - order is determined by annotation - test is not OD


darko-marinov:
Please split this big PR into smaller ones as per my email.

PR #852, opened 2023-10-24
Add detected flaky test

219sansim:
There is 1 ID flaky test in [pac4j](https://github.com/pac4j/pac4j) repo

Steps to reproduce:
Run the following command in the base of the repository -

```mvn edu.illinois:nondex-maven-plugin:2.1.7-SNAPSHOT:nondex```

Test fails on `nondexSeed=974622`

I verify it with the command:
`mvn -pl pac4j-oauth edu.illinois:nondex-maven-plugin:2.1.7-SNAPSHOT:nondex -Dtest=org.pac4j.oauth.profile.JsonHelperTests#testToJSONString -DnondexRuns=10`

Please find log files here: `~/pac4j/pac4jnondexlog.log`
VM name: ` fa23-cs527-054.cs.illinois.edu`

### DIFF
--- a/format_checker/forked-projects.json
+++ b/format_checker/forked-projects.json
@@ -300,6 +300,7 @@
   "https://github.com/orbisgis/h2gis": "unforked",
   "https://github.com/orbit/orbit": "unforked",
   "https://github.com/outbrain/aletheia": "unforked",
+  "https://github.com/pac4j/pac4j": "unforked",
   "https://github.com/paleozogt/symzip-plugin": "unforked",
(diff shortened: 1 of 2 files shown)


darko-marinov:
Please remove NetID from the public PR. How did you check that the test is `ID` and not `NOD`?

219sansim:
> How did you check that the test is `ID` and not `NOD`?

I ran the test with NonDex multiple times and it failed deterministically with the same nondex seeds. If a test is `NOD`, it may fail or pass with the same NonDex seed. That is how I concluded that it is an `ID` test.

PR #856, opened 2023-10-25
update secor flaky test type

bjchris32:
The `OD` status is caused by two other tests `testDelimitedTextFileWriter` and `testDelimitedTextFileReader` in the same test class.
To detect them, I printed some logs in the first line of each test and used NonDex to run it 10 times.
The result shows that when we run `testDelimitedTextFileWriter` or `testDelimitedTextFileReader` at first, all of the tests in the class will fail with the following error message.
```
[ERROR] Errors:
[ERROR] Test mechanism.Test mechanism
[ERROR]   Run 1:
Misplaced or misused argument matcher detected here:

-> at com.pinterest.secor.io.FileReaderWriterFactoryTest.mockSequenceFileWriter(FileReaderWriterFactoryTest.java:170)
-> at com.pinterest.secor.io.FileReaderWriterFactoryTest.mockSequenceFileWriter(FileReaderWriterFactoryTest.java:171)
-> at com.pinterest.secor.io.FileReaderWriterFactoryTest.mockSequenceFileWriter(FileReaderWriterFactoryTest.java:172)
-> at com.pinterest.secor.io.FileReaderWriterFactoryTest.mockSequenceFileWriter(FileReaderWriterFactoryTest.java:172)
-> at com.pinterest.secor.io.FileReaderWriterFactoryTest.mockSequenceFileWriter(FileReaderWriterFactoryTest.java:173)

You cannot use argument matchers outside of verification or stubbing.
Examples of correct usage of argument matchers:
    when(mock.get(anyInt())).thenReturn(null);
    doThrow(new RuntimeException()).when(mock).someVoidMethod(anyObject());
    verify(mock).someMethod(contains("foo"))

This message may appear after an NullPointerException if the last matcher is returning an object
like any() but the stubbed method signature expect a primitive argument, in this case,
use primitive alternatives.
    when(mock.get(any())); // bad use, will raise NPE
    when(mock.get(anyInt())); // correct usage use

Also, this error might show up because you use argument matchers with methods that cannot be mocked.
Following methods *cannot* be stubbed/verified: final/private/equals()/hashCode().
Mocking methods declared on non-public parent classes is not supported.

[ERROR]   Run 2:
Misplaced or misused argument matcher detected here:

-> at com.pinterest.secor.io.FileReaderWriterFactoryTest.mockSequenceFileWriter(FileReaderWriterFactoryTest.java:170)
```

I will open additional PR to fix the OD.


darko-marinov:
Please see email for why these tests are `OD`, and what about `Test mechanism.Test mechanism`.

bjchris32:
@darko-marinov , I had replied in campuswire discussion here: https://campuswire.com/c/G7A0E96FD/feed/473.
I am still wondering if I should delete the `Test mechanism.Test mechanism` or not.
Currently, I just leave it as it is.
Do you think I could delete it after our discussion on campuswire?

darko-marinov:
Yes, delete it. Did you check the rest of `pr-data.csv` for any similar case?

bjchris32:
> Yes, delete it. Did you check the rest of `pr-data.csv` for any similar case?

I checked the rest of `pr-data.csv` and only found one record in `secor` that should be deleted.
I've deleted it in my latest commit.
Thank you!

PR #897, opened 2023-11-03
Update pr-data.csv

prathyushreddylpr:


darko-marinov:
I'm accepting this PR, but can you follow up on the real PR and address modifiable collection, before you may be rejected?

PR #914, opened 2023-11-05
Update pr-data.csv

same8891:
These tests were deleted in commit 6fb2e547ff0d08be41de49a269ae3dd4c47c8149. The commit url is: https://github.com/apache/ratis/commit/6fb2e547ff0d08be41de49a269ae3dd4c47c8149.


darko-marinov:
Find the commit that deleted these tests, check they were actually deleted (not moved or renamed), and put the commit link in the last column (notes) separated with `;` from the existing link.

same8891:
I add the commit link in the last column (notes) separated with ; from the existing link.

PR #919, opened 2023-11-06
Added a script to identify type of test

kavvya97:
The script helps to determine the type of test (with approximation) as **OD, ID, NOD or not flaky**.  It should be placed at the root of the repository. It accepts the following arguments as command-line parameters.
<module_name>: Name of test module
<commit_hash>: Commit hash
<test_name>: Name of the test
<test_class>: Name of the test class

The script runs individual level tests, nondex test, nondex test with multiple nondex runs (-DnondexRuns=50), class-level tests and module-level tests. The script generates a <test_name> directory and creates a logs folder and dumps all the class-level, module-level, nondex runs in the logs folder.

The `verify_test_type.png` is a flowchart which helps to analyze the type of the test. By checking the logs files and analyzing the stored output and comparing with the flowchart, we will able to ascertain the type of test as **OD, ID, NOD or not flaky**


darko-marinov:
How does the approach differ from Fig. 3 in https://mir.cs.illinois.edu/marinov/publications/LamETAL20LongitudinalFlakyTests.pdf (page 12/29)?

kavvya97:
> How does the approach differ from Fig. 3 in https://mir.cs.illinois.edu/marinov/publications/LamETAL20LongitudinalFlakyTests.pdf (page 12/29)?

The approach we used is much simpler because we only run Nondex and simple `mvn test` at various levels. 
1. It does not IDflakies
2. OD tests are not classified as brittle, victim
3. The script does not run Nondex with the fixed failed seed

The flowchart explains the branches in the script. However, It is not as accurate and comprehensive as the diagram in the paper. We can consider removing it if required

darko-marinov:
Accepting this as it offers a good starting point for future students to work on.

PR #901, opened 2023-11-03
Two tests in apache/flink have had their modules renamed

krishnanand5:
The two tests in question are :

- `org.apache.flink.connectors.hive.HiveTableSinkTest.testWriteComplexType`.
- `org.apache.flink.connectors.hive.TableEnvHiveConnectorTest.testUDTF`.

As per the latest version of the apache/flink repository, both modules have been renamed to `HiveRunnerITCase` and `TableEnvHiveConnectorITCase` respectively.

The commit where the renaming of modules has occurred seems to be `850e3a520e55549a140138c81e433c44e641bdbc`.

Link to Commit - https://github.com/apache/flink/pull/12333/commits/850e3a520e55549a140138c81e433c44e641bdbc

PR #986, opened 2023-11-15
Marking 2 FLINK tests as DWF

krishnanand5:
With reference to https://github.com/TestingResearchIllinois/idoft/pull/901, the commit that was accepted here only had the rejected status updated for one test.

The other two tests for which issues were raised but the developer declined to have them fixed, have been marked as `DeveloperWon'tFix` in this PR.


darko-marinov:
Do we really need 8 commits for a one-line change?

PR #989, opened 2023-11-15
Dubbo/NIO test status change to DeveloperFixed

SaaiVenkat:
I ran the test `testNoInvoke()` multiple times (10 times) in the same JVM in the IDoFT commit, and the test failed. However, after the patch at commit [669705c7d3fa69244cb3301af0f2cd1360609857](https://github.com/apache/dubbo/commit/669705c7d3fa69244cb3301af0f2cd1360609857), the tests passed, when ran for more than 10 times in the same JVM.


darko-marinov:
Please move the `DeveloperFixed` commit link into the last column (`Notes`) not in the second-last (`PR Link`). Ideally, extend the checker to check for this case.

SaaiVenkat:
> Please move the `DeveloperFixed` commit link into the last column (`Notes`) not in the second-last (`PR Link`). 

I have updated the record Professor. 

> Ideally, extend the checker to check for this case.

I am sorry for not understanding properly. By `check for this case`, do you mean to update the checker so that `DeveloperFixed` commits are only in the last column and not present in the last before column?

darko-marinov:
Yes, for `check for this case` I meant what you said.

PR #1028, opened 2023-11-17
Spring boot/Fixed order status update

SaaiVenkat:
Based on the discussion with the author https://github.com/spring-projects/spring-boot/issues/38379, I have updated the status to fixed order


zzjas:
The code change doesn't match your PR description.

darko-marinov:
You need not ask developers if they want a fixed order when they already have a fixed order.

SaaiVenkat:
> You need not ask developers if they want a fixed order when they already have a fixed order.

Even though the order is fixed, the 'OD-Brit' test 'test002' fails when executed in a standalone manner, but passes when run the entire test class in run. I raised this issue to the developer and he said that it's expected to be order dependent. What should be the status of this test?

SaaiVenkat:
Professor @darko-marinov, I have updated the PR based on the discussion at https://campuswire.com/c/G7A0E96FD/feed/666

PR #1033, opened 2023-11-17
Auto run nondex

MyEnthusiastic:
Update to optimize the script


auto-run-nondex/runNondex.sh
```diff
@@ -16,7 +16,7 @@ runNondex () {
     input="modnames"
     while IFS= read -r line
     do
-        mvn edu.illinois:nondex-maven-plugin:$nondex_version:nondex -pl :$line -Dlicense.skip=true | tee ./.runNondex/LOGSSS/$line.log
+        mvn edu.illinois:nondex-maven-plugin:$nondex_version:nondex -pl :$line -Dlicense.skip=true -Drat.skip=true -DlicenseCheck.numUnapprovedLicenses=99999 -fae | tee ./.runNondex/LOGSSS/$line.log
```
darko-marinov, line 19: 1. Do the flags require explicit `=true`?

2. Which example needed that `numUnapprovedLicenses`?

3. Does the rest of script support the case where multiple modules have a test failure? (Maybe also expand `-fae` to the full name, so it's easier to read/maintain.)

MyEnthusiastic:
I also find the the mvn install is important for the single module --- but is un-necessary if we want to run all the modules.

It is need for the single module is because we might not have the dependency it need built, but for all modules the dependency will be automatically "compiled" first

auto-run-nondex/runNondexAll.sh
```diff
@@ -0,0 +1,5 @@
+# Assume this script is run from the folder of .pom.xml files
+
+mvn edu.illinois:nondex-maven-plugin:2.1.1-SNAPSHOT:nondex -Dlicense.skip=true -Drat.skip=true -fae | tee ./nondex.log
```
darko-marinov, line 3: You don't want `2.1.1-SNAPSHOT` (because 2.1.1 has been released), and you don't need `=true`. It'd be hard to review all other changes, which are rather big. Can you first propose a new PR with the minimal changes for the functionality you want to add?

darko-marinov:
Please squash your changes into one commit. Can your new script be just an option in the old script, to run on all modules or only one?

MyEnthusiastic:
> Please squash your changes into one commit. Can your new script be just an option in the old script, to run on all modules or only one?

I squashed them --- thanks for the suggestion ---  the old script can run all --- but it was very complex and need python dependency to convert to HTML and did many un-necessary works. The reason I add the new file is to provide a quick start :)

alexbadia1:
Hi, the status of this PR is a bit unclear. Are there any specific concerns or changes needed before this PR can be accepted?

This PR also seems to be a duplicate of https://github.com/TestingResearchIllinois/idoft/pull/1069/commits and it is unclear which is final.

auto-run-nondex/runNondex.sh
alexbadia1: 1. Do the flags require `explicit =true`?

The explicit `true` should not be required, but improves readability by making the purpose of the flags `-Dlicense.skip=true` `-Drat.skip=true` as boolean flags

2. Which example needed that `numUnapprovedLicenses`?

I don't know, but is a nice to have as a  workaround for a project with many dependencies that have unapproved licenses.

3. Does the rest of script support the case where multiple modules have a test failure? (Maybe also expand -fae to the full name, so it's easier to read/maintain.)

Changing -fae to --fail-at-end might be a good idea for readability

PR #1046, opened 2023-11-19
Update pr-data.csv

Rette66:
Delete non flaky test and update opened test


zzjas:
Checked the logs for the deleted test and it's not flaky at the reported commit so ok to delete.

For the `Opened` one, could you follow up with the developer? Also, if the test relies on a longer timeout to pass, then it's not `ID`. Can you confirm by running the test without NonDex? It should fail in some cases, making it `NOD`.

PR #1061, opened 2023-11-22
Changed status of `testHttpCode` test to DeveloperFixed

yesh385:
This PR changes the status of an NOD test named `testHttpCode` to DeveloperFixed. The commit which reduces the flakiness for this test is:
https://github.com/CrawlScript/WebCollector/commit/a41198989b84ab0e99ce821e6a752f57209c6991

This commit changes the `url` in `testHttpCode` from http://www.hfut.edu.cn/ch/ to https://www.bing.com/ which reduces the flakiness of this test since Bing is more accessible than the other website.


darko-marinov:
Please squash your 16 commits into 1. How does the change from one URL to another fixes the flakiness? Is it just that bing.com is accessible more often than the other website?

yesh385:
Yes professor. Bing is more accessible and returns 200 while the other website ofter returns 302.

darko-marinov:
So the flakiness was reduced but not resolved (because bing.com, or the network to it from the test runner, can also go down). Could the test mocked the network?

darko-marinov:
Maybe at least add a note that the flakiness was reduced not removed.

yesh385:
Hello professor,
The test is not mocking the network. It is making an actual HTTP request using the OkHttp library and returning a `Page` object representing the response. I will add a note mentioning that the flakiness is reduced and not entirely removed.

darko-marinov:
Yes, add a note just after the commit link ` reduces the flakiness`.

yesh385:
Thank you professor. Added the note.

darko-marinov:
Thanks for the note. Please squash your 5 commits into 1.

yesh385:
Done professor.

PR #1067, opened 2023-11-23
Added script for checking stale entries and sorting forked-projects.json

anirudh711:
The script works in unison with the rest of the scripts present in the idoft codebase. It is called when `format_checker/main.py` is executed. 
The script does the following on every run of `format_checker/main.py`:
- It runs a search on `pr-data.csv`, `gr-data.csv`, `py-data.csv` to check if there are any unused entries in forked-projects.json. If yes, it returns an error with the entries using the existing logging mechanisms in the codebase. If not, a success log is shown
- It sorts the data in alphabetical order 

This script can be extended to delete the stale entries in forked-projects.json rather than just showing the stale entries to be deleted. Do let me know if that needs to be done instead.

### DIFF
--- a/format_checker/forked_projects_util.py
+++ b/format_checker/forked_projects_util.py
@@ -1,0 +1,104 @@
+import json
+import os
+import csv
+import sys
+import errorhandler
+import logging
+from utils import (
+    log_info,
+    log_esp_error,
+    log_std_error,
+    log_warning
+)
+
+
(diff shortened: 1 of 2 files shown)


darko-marinov:
Please squash your 2 commits into 1.

I don't understand "It sorts the data" -- it should check if the data is sorted; it can't sort and update the data.

PR #1131, opened 2023-11-30
Change the testEquals2AnnotationsUnorderedParam test to be developerwont fix

FarmersWrap:
The developers declared that they won't fix this bug and they think this is due to the JVM.


zzjas:
Although the PR Link is the fix for another test, the test changed here is very similar so I think it's ok to mark it as `DeveloperWontFix` @darko-marinov

PR #1134, opened 2023-11-30
Opened pr and found new tests

lxb007981:
In `apache/jackrabbit-oak`: Opened a [pr](https://github.com/apache/jackrabbit-oak/pull/1220) fixing two flaky tests. For one of the tests there is already a pr opened but was closed by the developer in 2018. So I reopened a new one.
In `alibaba/spring-cloud-alibaba`: Opened a [pr](https://github.com/alibaba/spring-cloud-alibaba/pull/3498) fixing 3 OD flaky tests, and then found out the fix can apply to other 4 tests which are also OD flaky.


darko-marinov:
What made your fixes for `ID` tests also fix `OD` tests? It'd be quite an interesting case. Do you have logs showing `OD` tests failing before your changes (ideally just running polluter-victim pairs) and then passing after your changes?

lxb007981:
> What made your fixes for `ID` tests also fix `OD` tests? It'd be quite an interesting case. Do you have logs showing `OD` tests failing before your changes (ideally just running polluter-victim pairs) and then passing after your changes?

Those 3 confirmed flaky tests and 4 newly found flaky tests in `alibaba/spring-cloud-alibaba` are `OD`, not `ID`. There are no `ID` tests involved in this pr?

lxb007981:
> Sorry I misread your `fixing 3 flaky tests, and then found out the fix can apply to other 4 tests which are also OD flaky` as if you wrote `fixing 3 ID flaky tests`. You didn't write `ID`. (As a general rule, it may be better to open separate PRs or at least commits for different issues to make reviewing easier.)
> 
> Where are the logs that show those 4 tests failing before your changes but passing after your changes?

Since there are both junit 4 and junit 5 tests at the same time I can't simply run iDFlakies. I've created a [issue](https://github.com/alibaba/spring-cloud-alibaba/issues/3503) providing some details about the situation. I could add more details if needed.

darko-marinov:
You claim that you fixed 4 tests. How do you show that if you can't run the tool?

lxb007981:
> You claim that you fixed 4 tests. How do you show that if you can't run the tool?

Added a more detailed [issue](https://github.com/TestingResearchIllinois/idoft/issues/1136) to explain it.

darko-marinov:
Thanks for clarifying! So each test is a brittle and fails alone. In what order do they pass -- even before you change?

lxb007981:
> Thanks for clarifying! So each test is a brittle and fails alone. In what order do they pass -- even before you change?

If we run [com.alibaba.cloud.nacos.configdata.NacosConfigDataLocationResolverTest.testCommonPropertiesHasLowerPriority](https://github.com/alibaba/spring-cloud-alibaba/blob/6843cb06da903c39527705e07c16a0e17a06e873/spring-cloud-alibaba-starters/spring-cloud-starter-alibaba-nacos-config/src/test/java/com/alibaba/cloud/nacos/configdata/NacosConfigDataLocationResolverTest.java#L181) first, then other tests can pass. This is because in this test it set up some necessary environment properties.

PR #1174, opened 2023-12-04
Adding open prs to idoft

njain2208:


zzjas:
Are these tests ID or NOD (do they depend on implementation or on time)? I saw several fixes that are timing-related.

njain2208:
I believe that they are ID tests, as the error was in the handling of nanoseconds. For example the below error:

[2023-11-20T17:01:38.907444-06:00[America/Chicago]] but found [2023-11-20T17:01:38.907-06:00[America/Chicago]]

As One value has nanoseconds have nanoseconds and the other doesn't the test was failing so I changed both values in the same format.

njain2208:
I have attached the `mvn test` log files of all the tests in the morpheus-logs folder in the VM.

PR #1185, opened 2023-12-05
Added 5 orika ID tests

FarmersWrap:
All tests passed normal "mvn test", 30 times nondex tests failed, one nondex seed generate one result(pass or fail).
Tests logs can be found here: /home/yu129/new_projects2/orika and /home/yu129/new_proj2_tests/orika/-logs


zzjas:
Ok to merge @darko-marinov

PR #1199, opened 2023-12-07
Update pr-data.csv [apache/hadoop]

yijut2:
Opened PR in apache/hadoop


zzjas:
Ok to merge @darko-marinov

PR #1090, opened 2023-11-27
Add script to run single test

bbelide2:
Added script to run a single test multiple number of times to determine the category of the test (ID or NOD).

PR #1240, opened 2024-04-26
Polished Script to quickly check ID or NID

average-avancha:
Polished script intended to run single runs of NonDex to quickly gauge if the flaky test is likely ID or NID.

[#1090](https://github.com/TestingResearchIllinois/idoft/pull/1090)


darko-marinov:
How does this functionality differ from https://github.com/TestingResearchIllinois/idoft/tree/main/verify_test_type ?

average-avancha:
Looks like [the following script](https://github.com/TestingResearchIllinois/idoft/tree/main/verify_test_type) would be superset of the functionality of this script and does a more rounds of checks at the project level to determine test type. 

The tradeoff for more through checks is longer runtime, while this script should be a quick check to determine likely ID or NID. If likely NID, then that script could be the next step in determining type.

darko-marinov:
How hard is it to change that `verify_test_type` to do fewer checks?

average-avancha:
It shouldn't be too difficult to make those changes in the script assuming the user understands functionality (the purpose of each level of checks to narrowing down the likely root cause of the failed tests). Ideally, they can just comment out the levels of tests they want to skip.

average-avancha:
Closing this PR as this script would duplicate work from: https://github.com/TestingResearchIllinois/idoft/tree/main/verify_test_type

PR #1259, opened 2024-10-04
Identified one fixed test and two moved tests

williamlin0825:
**The fixed test:**

**org.wildfly.extension.undertow.UndertowSubsystemTestCase.testRuntime:**
Here is the commit that fixed this flaky test:
https://github.com/wildfly/wildfly/commit/7bbf48d473f442b690f3696172d368ab51bd7025

The Nondex logs from before and after bug fix is at /home/jhihwei2/logs.

**The moved tests**

**org.jboss.as.test.smoke.mgmt.datasource.AddMySqlDataSourceOperationsUnitTestCase.testAddDsAndTestConnection:**
The AddMySqlDataSourceOperationsUnitTestCase class is renamed to AddDataSourceOperationsUnitTestCase but the testAddDsAndTestConnection method is remained the same. The new row is added at 6417.

**com.google.cloud.teleport.templates.common.BigQueryConvertersTest.testFailsafeJsonToTableRowValidInput:**
The test had been moved from /. to /v1 module and the test path and test method body remain the same in the module. The new row is added at 4370.


Ajitesh27:
LGTM

PR #1264, opened 2024-10-08
Add 2 PR's, 1 deletion

Mukta13:
### For test name:
**1. org.apache.nifi.minifi.toolkit.configuration.ConfigMainTest.testTransformRoundTripNestedControllerServices:**

ConfigMainTest does not exist anymore. Deleted at commit id:3aa02a022e154416cc6f69acc327aeb26f2a89b2. Changed
the status to deleted in this pr.

**2. org.apache.flink.table.planner.plan.hints.stream.StateTtlHintTest.testStateTtlHintWithJoinHint:**

Opened the pull request. Changed the status to Opened in this pr and have also attached the pr link in it.

**3. com.google.common.collect.testing.testers.MapToStringTester.testToStringWithNullKey[Synchronized.setMultimap [collection size: several].asMap [collection size: several]] :**

The module name guava-test does not exist anymore in the latest repo. This has been moved to guava-testlib
instead. Moved here:
[commit id](https://github.com/google/guava/commit/ae2c37b1ba5de61df02e1f018e294f255c071c32)

**4. org.drools.model.codegen.execmodel.PropertyReactivityTest.testUnwatchWithFieldBindingAndMvel:**

Opened the pull request. Changed the status to Opened in this pr and have also attached the pr link in it.

PR #1266, opened 2024-10-09
Changed to Dev wont fix for guava

Mukta13:
Continuing the conversation from [https://github.com/TestingResearchIllinois/idoft/pull/1264](https://github.com/TestingResearchIllinois/idoft/pull/1264):

If you look at the repo from this commit (the commit when the test was reported )- (https://github.com/google/guava/tree/a197d99b96d69ea75a2c4a15ecd8a730dd3270d6): [commit tree]

The module guava-test is not available, and if I try to find the MapToStringTester.java file in this, it clearly shows that it's only present under guava-testlib module and not under guava-test. The MapToStringTester.java was indeed moved under guava-testlib in this commit ae2c37b1ba5de61df02e1f018e294f255c071c321 and it hasn't been moved from it after that.

Edit: Changed to dev wont fix


Ajitesh27:
Please go through the instructions to add Moved/Renamed status in README. The old line must have the status MovedOrRenamed along with the notes. You must then add a new line for the moved test with latest details.

FAQ
How to indicate that a test is renamed or moved to a different location?
If the test method body is the same between two versions (e.g., if in old_sha, some.test.name has the same test method body as some.other.test.name in new_sha), we consider the two different versions of the test to be the same test. For the row with the older sha, please change the Status to MovedOrRenamed and add [Notes](https://github.com/TestingResearchIllinois/idoft/tree/main#adding-notes) describing which version the test is found to be renamed/moved.

If the test method body is different, we consider the two different versions of the test to be two different tests. For the row with the older sha, please change the Status to Deleted and add [Notes](https://github.com/TestingResearchIllinois/idoft/tree/main#adding-notes) describing which version the test is found to be renamed/moved and how the test method body differs between the two versions.

For either case, please also add a new row for the newer sha and test name. Once the preceding changes are made, all future pull request updates should only be made to the row with the newer sha.

Ajitesh27:
You can clean up the fully qualified path and also I think the test is marked with @Ignore

Mukta13:
> You can clean up the fully qualified path and also I think the test is marked with "@ Ignore"

I have made the changes, please let me know if it looks fine or not.

Thanks!

Mukta13:
Changed to Dev wont fix

darko-marinov:
Please squash commits and keep only` DeveloperWontFix`.

PR #1272, opened 2024-10-09
Open two PRs and report one NameOrReport

SijieMei:
Open two PRs
report one NameOrReport for nifi. The new test required java 21 or above to build.

### DIFF
--- a/pr-data.csv
+++ b/pr-data.csv
@@ -6148,7 +6148,7 @@ https://github.com/uniVocity/univocity-parsers,7e7d1b3c0a3dceaed4a8413875eb1500f
 https://github.com/uniVocity/univocity-parsers,7e7d1b3c0a3dceaed4a8413875eb1500f2a028ec,core,com.univocity.parsers.common.processor.MultiBeanListProcessorTest.testMultiBeanProcessor,ID,,,
 https://github.com/uniVocity/univocity-parsers,7e7d1b3c0a3dceaed4a8413875eb1500f2a028ec,core,com.univocity.parsers.common.routine.AbstractRoutinesTest.testIterateJavaBeans,ID,,,
 https://github.com/uniVocity/univocity-parsers,7e7d1b3c0a3dceaed4a8413875eb1500f2a028ec,core,com.univocity.parsers.common.routine.AbstractRoutinesTest.testParseAllJavaBeans,ID,,,
-https://github.com/uniVocity/univocity-parsers,7e7d1b3c0a3dceaed4a8413875eb1500f2a028ec,core,com.univocity.parsers.common.routine.AbstractRoutinesTest.testWriteAllJavaBeans,ID,,,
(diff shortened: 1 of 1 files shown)


Ajitesh27:
Please go through the instructions to add Moved/Renamed status in README. The old line must have the status MovedOrRenamed along with the notes. You must then add a new line for the moved test with latest details.



FAQ
How to indicate that a test is renamed or moved to a different location?
If the test method body is the same between two versions (e.g., if in old_sha, some.test.name has the same test method body as some.other.test.name in new_sha), we consider the two different versions of the test to be the same test. For the row with the older sha, please change the Status to MovedOrRenamed and add [Notes](https://github.com/TestingResearchIllinois/idoft/tree/main#adding-notes) describing which version the test is found to be renamed/moved.

If the test method body is different, we consider the two different versions of the test to be two different tests. For the row with the older sha, please change the Status to Deleted and add [Notes](https://github.com/TestingResearchIllinois/idoft/tree/main#adding-notes) describing which version the test is found to be renamed/moved and how the test method body differs between the two versions.

For either case, please also add a new row for the newer sha and test name. Once the preceding changes are made, all future pull request updates should only be made to the row with the newer sha.

pr-data.csv
```diff
@@ -4369,7 +4369,7 @@ https://github.com/GoogleCloudPlatform/DataflowTemplates,5094c7b39de511c9ed441d9
...
 https://github.com/GoogleCloudPlatform/DataflowTemplates,cc17ed154d8c3c385ab3f5a3653a6e4c02539bfa,v1,com.google.cloud.teleport.templates.common.BigQueryConvertersTest.testFailsafeJsonToTableRowValidInput,ID,,,
 https://github.com/GoogleCloudPlatform/DataflowTemplates,5094c7b39de511c9ed441d9fde28553a88f68e4b,.,com.google.cloud.teleport.templates.common.DatastoreConvertersTest.testCheckNoKeyAllInvalid,ID,,,
-https://github.com/GoogleCloudPlatform/DataflowTemplates,5094c7b39de511c9ed441d9fde28553a88f68e4b,.,com.google.cloud.teleport.templates.common.DatastoreConvertersTest.testCheckNoKeyBothCorrectAndInvalid,ID,,,
+https://github.com/GoogleCloudPlatform/DataflowTemplates,0d01accc1cfa29b762a764bc226496bfde25165c,v1,com.google.cloud.teleport.templates.common.DatastoreConvertersTest.testCheckNoKeyBothCorrectAndInvalid,ID,Opened,https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1919,
```
darko-marinov, line 4372: Please do not change the commit SHA where the flaky was reported. What's the new SHA that you want to add? Maybe you can report in between that the test was moved. You could get +2 points for that (but need to find the commit where the test was moved).

SijieMei: I removed the change of this line and will figure out the correct format to create a new PR. Sorry for the confusion.

darko-marinov:
Please squash your 4 commits into 1.

SijieMei:
Hi Professor, I squashed my commits into one.

PR #1273, opened 2024-10-09
Opened 1 PR for 2 tests

everbrightw:
Opened 1 PR: [https://github.com/apache/incubator-kie-drools/pull/6119](https://github.com/apache/incubator-kie-drools/pull/6119)

For tests: 
```
org.drools.model.codegen.execmodel.AccumulateTest.testInlineAccumulateWithAnd
org.drools.model.codegen.execmodel.AccumulateTest.testBindingOrderWithInlineAccumulateAndLists
```


Ajitesh27:
LGTM

everbrightw:
Hi Darko @darko-marinov Ajitesh @Ajitesh27 

I simplified the issue happening in the incubator-kie-drools project and why it is failing under NonDex ONE.


```java
Set<String> set = new HashSet<>();
set.add("apple");
set.add("orange");
set.add("banana");

// Iteration 1: Using Iterator
System.out.println(set);

// Iteration 2: Using Stream.forEach()
set.stream().forEach(e -> System.out.println(e));

```
### Some Context:

The project's business logic relies on these two types of iteration order being ‘relatively deterministic,’ and to quote the maintainer

> 2 iterations of the same unchanged HashSet instance in the same JVM execution

I tried to run NonDex in mode ONE and error is still occurring. 

### NonDex Behavior:

When running NonDex in ONE mode on the simplified example:
- Iteration 1 Iterator: The order is shuffled, and if I added another System.out.println(set) after, the order remains the same in each execution (which is not the case under FULL mode).
- Iteration 2 Stream.forEach(): The order of the `Stream.forEach()` iteration remains consistent across all runs of NonDex.

And since the Iteration 1 is shuffled and Iteration 2 remains consistent, the error occurred in their business logic. 

### What I Searched:

- I read the Stream API doc [here](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html#forEach(java.util.function.Consumer)) I am confused by 

> 'The behavior of this operation is explicitly nondeterministic. For parallel stream pipelines, this operation does not guarantee to respect the encounter order of the stream,...' 

- and a post answered by a Java library engineer: https://stackoverflow.com/questions/34247318/does-stream-foreach-respect-the-encounter-order-of-sequential-streams

- and a filed issue related to the order of stream.forEach: https://bz.apache.org/netbeans/show_bug.cgi?id=257129

Specifically, I do not know if, for a HashSet, **during the same execution**, and **HashSet remains unmodified**, whether stream.forEach preserved the same 'order' as the Iterator does for **one** execution. 

### Questions:

1. Is this an issue with NonDex in this scenario? 

2. Or, Could it be that `Stream.forEach`'s inherent implementation is very different from Collection's `Iterator`, and while `HashSet`'s `Iterator` is not deterministic across different runs, `Stream.forEach` can not be assumed to hold the same order with `Iterator` even during one execution. In this case, for the PR, relying on 'matching' the order between these two iteration on a HashSet could be a problem.

PR #1289, opened 2024-10-10
Updated DeveloperFixed test for alibaba-druid

yugvajani:
Repo - https://github.com/alibaba/druid
Test - com.alibaba.druid.bvt.pool.DruidDataSourceTest_exceptionSorter.test_event_error
Developer Fixed Commit - https://github.com/alibaba/druid/commit/b2fcc80e2ef70c7abf08676426d933569645897f

### DIFF
--- a/pr-data.csv
+++ b/pr-data.csv
@@ -642,7 +642,7 @@ https://github.com/alibaba/alibabacloud-tairjedis-sdk,f6215930b9e21a5252c95296d9
 https://github.com/alibaba/alibabacloud-tairjedis-sdk,f6215930b9e21a5252c95296d9924b3c4524d9f6,.,com.aliyun.tair.tests.tairzset.TairZsetTest.zrevrank,NOD,,,
 https://github.com/alibaba/alibabacloud-tairjedis-sdk,f6215930b9e21a5252c95296d9924b3c4524d9f6,.,com.aliyun.tair.tests.tairzset.TairZsetTest.zscore,NOD,,,
 https://github.com/alibaba/asyncload,9891e7086bff81ebdd4991435d21b549e48e874d,.,com.alibaba.asyncload.AsyncLoadThreadLocalTest.testInheritableThreadLocalRunnerSet,OD-Vic,Unmaintained,,last commit on 2018-05-15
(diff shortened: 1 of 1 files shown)


darko-marinov:
What evidence do you have that `b2fcc80e2ef70c7abf08676426d933569645897f` fixed the test? Do you have logs from running right before and right after this commit?

yugvajani:
Found the fixed commit using the git blame for file. Not being able to run the nondex build for the earlier commit as build fails due to jdk 1.6 version issue.

darko-marinov:
I'm not convinced that this commit fixed this test. I don't understand the comment about 1.6. You'd need a stronger justification.

yugvajani:
The nondex tool does not work for jdk 6 and the commit with the fix was using jdk 6 . So when i try to build the project and run nondex command it gives me an error as nondex does not support jdk 6.

darko-marinov:
How can you claim that this specific commit fixes the test if you can't compile and run the code in that commit? Also, is the commit that you're adding in the notes after the commit that was reported? I don't remember if NonDex ever supported Java 6. Maybe you can try an older version of NonDex, e.g., `edu.illinois:nondex-maven-plugin:1.0.0`.

darko-marinov:
Any progress on this case? You can even change `pom.xml` to see if the project could compile and run tests with Java 8.

yugvajani:
I am trying to figure out the commit which might have fixed it using git bisect as the one mentioned in the pr seems to be not fixing it .

darko-marinov:
For bisect, you may consider the script from https://github.com/TestingResearchIllinois/idoft/tree/main/auto-check-fix-commit  It may help you go faster, or it may slow down (or it may make you equally fast, so I fully cover all the cases :).

darko-marinov:
I'm still waiting for a reply here. We can close this PR for negative "points", or you can provide more info (especially logs). In particular, if you can learn how to run `git bisect`, you may even present that in the class. You can ask @wtrzas2 for some help.

yugvajani:
I had discussed with you this in class where i tried running auto check
commit but it didn’t work for me. I am fine with closing the pr as i spent
some time on it but was not able to get anything.

PR #1300, opened 2024-10-11
1 Accepted

AmitPr:


darko-marinov:
Please separate the two changes into two PRs. We need to discuss more the case where someone else opened a PR.

AmitPr:
Opened a second PR, #1301, as requested

darko-marinov:
Please squash 3 commits into 1.

Don't keep your work on `main` as you'll run into problems.

PR #1306, opened 2024-10-13
Adding Rejected status for guava PR

hermya:
Adding Rejected status for guava PR
https://github.com/google/guava/pull/7387

Rejected reason:
- Guava has **its own jdk** for testing, which **randomizes the seed used for hash-ordering**.
- They have observed **different hash-ordering** when the **seed changes**, but **not** using the **same VM**.  <br> (Basically, in their testing jvm, hash-ordering won't change for result of `map.toString()` method called after `map.toString()` method)
- They are okay with taking the risk of not fixing test for now. They may consider working on such tests if it fails in 'practice' (without nondex)


darko-marinov:
Please revise or remove the note, per email.

PR #1322, opened 2024-10-16
RepoArchived for 4 flaky tests

mumbler6:


darko-marinov:
Good. Can you run some script to find all archived repos? You can find some scripts in the IDoFT repo directories named `auto-*`: https://github.com/TestingResearchIllinois/idoft

mumbler6:
I ran the script on gr-data.csv and pr-data.csv, there are no archived repos there. However, the script doesn't work on py-data csv because the columns are different.

darko-marinov:
Can you change the script to work on `py-data.csv` as well? We can always give some "points" for that.

mumbler6:
I can fix this, but I've noticed that the code to detect Archived repositories is not working. Currently the code is looking through the html response:

```
        if "This repository has been archived by the owner." in r.text:
            archived.append(url)
            print("archived: ", url)
 ```
 
But this isn't working on repositories that I've found as archived, such as https://github.com/epsagon/epsagon-python. The HTML response doesn't contain the divs that mark it as archived. 
 
 I thought about using the requests library to make API requests directly to GitHub, but it seems like this method is rate limited, which will sometimes cause 403 status codes for all repositories. 
 
 I will change the script to work on `py-data.csv` for now, let me know what you think about this issue.

PR #1332, opened 2024-10-16
developer wont fix incubator as per class announcement

Mukta13:
added developer wont fix to the assigned incubator test as per class announcement


darko-marinov:
We're still discussing the actual cause. You're welcome to join @everbrightw I'll leave this PR pending until the issue of `HashMap::spliterator()` is not resolved.

darko-marinov:
This test is in a different module, so you or @everbrightw should investigate whether it's due to the same issue (consistent order between regular `iterator()` and stream iterator).

everbrightw:
@darko-marinov  I can confirm this is because of the same issue.

darko-marinov:
Thanks for confirming. I'm accepting this PR.

PR #1339, opened 2024-10-17
Update pr-data.csv with Opened PR for gcp-ingestion

ZhanwangZhou:
The issue occurs due to the non-deterministic behavior of `IOFunction.unchecked(Json::asString)`. It converts `PubsubMessage` objects, which contain `Map<String, String>`, into strings. The Map to String convertion in Java is non-deterministic, so the Strings in `actual` are in random order and may not equal to those in `expectedFromLite`.


darko-marinov:
The addition looks good, but the format checker fails: `Invalid Project URL: "https://github.com/mozilla/gcp-ingestion", Reason: Please add the new project to format_checker/forked-projects.json`. You can see GitHub Action results. The specific link is https://github.com/TestingResearchIllinois/idoft/actions/runs/11378550327/job/31654515951?pr=1339

PR #1360, opened 2024-10-22
opened pr for two flaky tests

mumbler6:
Opened PR for two flaky tests. The flaky tests are the same, just duplicates in two different file paths. My PR fixes both. I didn't open two PR's because I wasn't sure if that would annoy the developers, as they are essentially the same test case.


darko-marinov:
Fine to open one PR for two tests. I'm not sure that real PRs need images, maybe skim https://meta.stackoverflow.com/questions/285551/why-should-i-not-upload-images-of-code-data-errors

PR #1367, opened 2024-10-22
Identified DeveloperFixed tests in vertx-mqtt

williamlin0825:
Found 5 fixed tests in vertx-mqtt.
The logs of before and after fix are at /home/jhihwei2/logs.
Since they are NOD tests, I run with DnondexRuns=300 to validate NOD fix.


darko-marinov:
Your `before` logs are much shorter than `after` logs, and some `before` logs appear incomplete (check the last few lines). Please double check that everything was run appropriately. Did you run `mvn clean` in between building the two SHAs?

darko-marinov:
@williamlin0825 please let @ajitesh27 know when you update the logs so he can check them.

darko-marinov:
I'm accepting this PR, but you should describe better to the developer how the tests are order dependent (and you need not refer to iDFlakies at all -- it's irrelevant how you found that the tests are `OD` but just show how running them in one order they pass but running them in another order they fail).

PR #1377, opened 2024-10-23
update dev-fixed for wildfly

everbrightw:
Update DeveloperFixed status for wildfly

before this fix, tests are failing, logs at:
`/home/yusenw2/wildfly/test_fail_before_pr_2.log`

after this fix, tests are passing, logs at:
`/home/yusenw2/wildfly/test_success_after_pr_2.log`


darko-marinov:
The code change doesn't look that related. Did you run NonDex for 10 or 100 seeds after the change?

everbrightw:
I did not dive deep why this fix works. I looked into the tests and they look fine to me. They eventually called an external library [wildfly-glow.](https://github.com/wildfly/wildfly-glow) I think this has to do with how their internal logic work or XML config for setup? 
I run with DnondexRuns=10, I can try 100 and update here.

everbrightw:
tests passed with running NonDex for 100 seeds after the change
log: `/home/yusenw2/wildfly/run_100_after_pr.log`

everbrightw:
finding the fix credit to this PR https://github.com/TestingResearchIllinois/idoft/pull/1350 on IDOFT
I was only testing and verifying for tests assigned to me :)

darko-marinov:
Thanks for these. Please feel free to completely focus on `forEach`, no need to add more of the existing tests. If `forEach` is supported well, it could find many new tests.

PR #1382, opened 2024-10-23
Updated pr-data.csv with DeveloperWontFix tests for incubator-kie-drools

yugvajani:


darko-marinov:
Why open a new PR rather than update the existing PR?

yugvajani:
I do not have an open PR for this

PR #1390, opened 2024-10-24
Corrected test paths

Ethan-Chin:
Corrected 3 test paths for `bytedeco/javacpp`


darko-marinov:
Thanks for the correction. Did these tests fail in the class constructor (and someone reported the name somewhat incorrectly)?

Ethan-Chin:
Sorry, I don't quite understand what you mean. These 3 tests were assigned to me and Maven cannot run these tests using the incorrect test name.

darko-marinov:
Your test names have no method names just class names. Why are there no method names? (It could be because the test classes crash in the constructor in some `@BeforeClass` step.)

Ethan-Chin:
The test won't crash with test names. I'm not sure why the three lines did not include test name in the first place, but the test paths are wrong anyway.

darko-marinov:
What are the correct "test paths" (fully qualified class names)? Were the test classes failing in the static initializer, class constructor, or some `@BeforeClass`?

darko-marinov:
I'm still waiting for a reply here. We can close this PR for negative "points", or you can explain where the tests fail.

Ethan-Chin:
> What are the correct "test paths" (fully qualified class names)? Were the test classes failing in the static initializer, class constructor, or some `@BeforeClass`?

No, the tests won't fail anywhere with the correct paths as you can see in my file diff:
```shell
org.bytedeco.javacpp.BufferTest
org.bytedeco.javacpp.EnumTest
org.bytedeco.javacpp.PointerTest
```

Those correct paths do not include the method names just because the original test paths do not include them in the first place, but it won't have problems if method names were added.

Ethan-Chin:
> What are the correct "test paths" (fully qualified class names)? Were the test classes failing in the static initializer, class constructor, or some `@BeforeClass`?

Apologies for the misunderstanding.

Yes. `org.bytedeco.javacpp.PointerTest`, `org.bytedeco.javacpp.EnumTest`, and `org.bytedeco.javacpp.BufferTest` all fail in their setUpClass() when using NonDex. This is the reason why these tests do not include specific method names.

darko-marinov:
Thanks for clarifying. Can you tell me here a direct link to `setUpClass` on GitHub (with that `#LineNumber` at the end)? At least that should be added in the `Notes` column, and maybe some (the first?) test method name should be provided, even if the test class can't even reach that method.

Ethan-Chin:
These are the links to `setUpClass`

```shell
https://github.com/bytedeco/javacpp/blob/master/src/test/java/org/bytedeco/javacpp/BufferTest.java#L78
https://github.com/bytedeco/javacpp/blob/master/src/test/java/org/bytedeco/javacpp/EnumTest.java#L92
https://github.com/bytedeco/javacpp/blob/master/src/test/java/org/bytedeco/javacpp/PointerTest.java#L69
```

darko-marinov:
Thanks for all the updates. The lines look much better now. Just one final request: replace the URLs like https://github.com/bytedeco/javacpp/blob/master/src/test/java/org/bytedeco/javacpp/BufferTest.java#L78 with URLs to the specific commit (otherwise as `master` changes, the line number may change and not be 78 as in this example).

Ethan-Chin:
Thanks. I have updated them with the commit id

PR #1403, opened 2024-10-27
tests marked as repoArchived in pr-data.csv and gr-data.csv

mumbler6:
Marked 105 tests in pr-data.csv as repoArchived,  2 tests in gr-data.csv

Some tests under `https://github.com/alibaba/fastjson` had issues / pr's linked in their notes section, I replaced these with `repoArchived" since these links were either 404 not found or closed.


darko-marinov:
Please undo changes on `alibaba/fastjson` as those would require more inspection. Keep in this PR only the changes that replace no status (blank string) with `RepoArchived`.

PR #1449, opened 2024-11-07
Update pr-data.csv with one DeveloperFixed test for tomcat_exporter

Qi-Cui:
Evidence has been sent via email.


darko-marinov:
Please move the commit from the second last column to the last column. Can you modify `format_checker` to automatically check for these cases?

@Ajitesh27 let me know when you check the evidence sent by email.

Qi-Cui:
> Please move the commit from the second last column to the last column. Can you modify `format_checker` to automatically check for these cases?
> 
> @Ajitesh27 let me know when you check the evidence sent by email.

Done. Yes I will modify the `format_checker` in a separate PR.

Ajitesh27:
LGTM

PR #1450, opened 2024-11-07
Update unmaintained status to py-data and minor fix to script

zhihao11ui:


py-data.csv
darko-marinov, line 1: It looks as if you changed each and every line here. Was there some issue with the EOL?

zhihao11ui: Thanks, I fixed the issue and that is exactly why

darko-marinov:
Change ` | 2020/01/17` to `;2020/01/17` (although even better would be `latest commit 2020/01/17`). Did you have any case where a test already has some non-blank status (e.g., `Opened`) but you wanted to add `Unmaintained`?

zhihao11ui:
For any non-blank status, I just skip them

darko-marinov:
Fine to skip non-blank. Please change those vertical bars to semicolons in all rows.

darko-marinov:
For `mojeek` you removed double quotes. Please return them back (or even better open first one PR to remove those double quotes and the TAB character that shouldn't have been there anyhow).

darko-marinov:
I see people working on unmaintained repos, so let me accept your PR even with its small mistakes. We can clean it later.

PR #1456, opened 2024-11-07
Dev wont fix - rejected

Mukta13:
This PR was opened way before it was announced that all the tests in this incubator repo will be marked as Dev Wont fix. This PR got rejected for the same reason as the prev one. Updating it here for the cumulative progress.


darko-marinov:
Thanks for reporting even this case despite initial negative "points".

PR #1471, opened 2024-11-10
3 tests that are not flaky

mumbler6:
I found 3 tests given to me that are not flaky. I ran nondex with 100 runs, and no failures at the specified commit (9641a013665e7012941b2362e9cba64fac4ba0e6). The log file is at `/home/bg16/ignite-3/modules/storage-page-memory/.nondex`


darko-marinov:
Did you run just the test method or the entire class(es) or the entire module? It's hard to tell just from the directory without the stdout log. The tests are likely `NOD` and failed for someone running NonDex in some run but are not `ID`.

mumbler6:
I just ran nondex on both test classes with no failures. If they are OD tests, I'm not sure on how to fix them. I've had trouble running the OD test tools.

darko-marinov:
For these tests, can you run NonDex on the entire module? You can even run `mvn test` multiple times, e.g., `for i in {1..10}; do mvn test |& tee LOG-${i}; done`.

For OD issues, you'd need to report what exactly you ran and how so someone can help debug the issues. We'll also have a presentation about OD tests this Thursday.

darko-marinov:
Any progress on finding if the tests may be `NOD`?

mumbler6:
I ran `mvn -pl modules/storage-page-memory edu.illinois:nondex-maven-plugin:2.1.7:nondex -DnondexRuns=20` and `for i in {1..10}; do mvn test -pl modules/storage-page-memory |& tee LOG-${i}; done` at commit sha 9641a013665e7012941b2362e9cba64fac4ba0e6.

The NonDex runs all passed, and after looking in each of the LOG-i files, it seems normally running mvn test passed as well. 

image

PR #1481, opened 2024-11-13
3 Skipped for undertow

Derrick2000:
In `JsrWebSocketServer07Test`, the method `testErrorHandling` is not marked with `@org.junit.Test`. Instead, there's a line of comment on top: 
image

I believe authors of this repo is currently excluding this test case, so resolving its flakiness seems unnecessary.

Also, the same goes to 
`JsrWebSocketServer08Test#testErrorHandling`
`JsrWebSocketServer13Test#testErrorHandling`
where both of the test class extend `JsrWebSocketServer07Test` without overriding `testErrorHandling`


Derrick2000:
> `Skipped` status is fine, but per the IDoFT README: `To use this status, please provide some [Notes](https://github.com/TestingResearchIllinois/idoft#adding-notes) on why the test should be skipped`. You should put a link to that line `FIXME` rather than posting an image.

Updated

PR #1488, opened 2024-11-14
Update auto-check-fix-commit README.md git bisect run process

wtrzas2:


auto-check-fix-commit/README.md
```diff
@@ -17,12 +17,19 @@ Reference - https://git-scm.com/docs/git-bisect/
 
 ### Usage
 
-1. Copy scripts `git-bisect-runner.sh` and `git-bisect-script.sh` to the directory of the project containing the fixed flaky test.
+1. Fork/clone this repo with the auto-check-fix-commit module containing the `git-bisect-runner.sh` and `git-bisect-script.sh` scripts.
 
+2. Within `git-bisect-runner.sh`, modify the `git bisect run ./git-bisect-script.sh $test_module $test_case` line to specify your path to `git-bisect-script.sh`, adding in `nohup` command at the beginning of the command to keep process running even after exiting shell/terminal.
```
darko-marinov, line 22: It shouldn't be necessary to modify the path in `git-bisect-runner.sh` -- if both that script and `git-bisect-script.sh` are in the same directory, then `git-bisect-runner.sh` should find its own directory and run `git-bisect-script.sh` from there. (The names for the scripts are not very good, but let's keep them as they are for now.)

Also, it shouldn't be necessary to add another `nohup` to this script if the first one is already run with `nohup`.

auto-check-fix-commit/README.md
```diff
@@ -17,12 +17,19 @@ Reference - https://git-scm.com/docs/git-bisect/
...
 
-2. Run the script using the below command
+3. Within `git-bisect-script.sh`, modify the build line `mvn clean install -pl $test_module -am -DskipTests -Dlicense.skip` to be the build command for your project with the DeveloperFixed test. Also make sure that the NonDex version, within the script, is correct for your specific project.
```
darko-marinov, line 24: Can this also be achieved without modifying the script itself but just by passing in some additional arguments for the `mvn` command(s)? For example, could one run with `git-bisect-runner.sh --flaky ... --fixed ... --module ... -- test ... --mvn-install "-Dspotless.skip -Denforcer.skip"`.

auto-check-fix-commit/README.md
```diff
@@ -17,12 +17,19 @@ Reference - https://git-scm.com/docs/git-bisect/
...
+
+4. Within the project containing the DeveloperFixed test, run the script using the command below. Modify arguments to path to script, specific flaky commit, fixed commit, module path, and test case of your project. This command will also ensure standard error and output messages go to `git_bisect_output.log`, while running command in the background.
 ```shell
-./git-bisect-runner.sh --flaky e6d76803f27133d7700811585f5310470e50e487 --fixed 5828d56a824748e7c91076842fad75efb42f92f9 --module core/server/worker --test alluxio.worker.block.BlockLockManagerTest#lockAlreadyReadLockedBlock
```
darko-marinov, line 25: It's good to leave some concrete example in addition to the general invocation.

auto-check-fix-commit/README.md
```diff
@@ -17,12 +17,19 @@ Reference - https://git-scm.com/docs/git-bisect/
...
+nohup /path/to/script/git-bisect-runner.sh --flaky <FLAKY_COMMIT> --fixed <FIXED_COMMIT> --module <MODULE_PATH> --test <TEST_CASE> > git_bisect_output.log 2>&1 &
```
darko-marinov, line 28: Maybe replace `> git_bisect_output.log 2>&1` with `&> git_bisect_output.log` for a modern Bash.

darko-marinov (changes requested):
Maybe the scripts should be changed somewhat so that the steps are easier (e.g., no need to change the scripts to run them but just allow passing in more arguments).

auto-check-fix-commit/README.md
wtrzas2: Got it! I will remove this step.

auto-check-fix-commit/README.md
wtrzas2: I believe it could. I can add: `'--mvn-install') shift; mvn_options="$1" ;;` after line 9 in `git-bisect-runner.sh` and then also in line 35 add at the end of the run command add `"$mvn_options"` for it to be: `git bisect run ./git-bisect-script.sh "$test_module" "$test_case" "$mvn_options"`.

And then in turn, `git-bisect-script.sh` would have to change mvn clean install line toI
`mvn clean install -pl $test_module -am -DskipTests -Dlicense.skip $mvn_options` and then also add `$mvn_options` to NonDex line. I can try testing this out on another branch.

auto-check-fix-commit/README.md
wtrzas2: Noted. Added concrete example back in.

auto-check-fix-commit/README.md
wtrzas2: Got it!

wtrzas2:
Committed and pushed most of the proposed changes for now, but am going to test out modifying the scripts to add in possibly accepting arguments with mvn command. Will comment again on this when done.

wtrzas2:
Changes made within scripts as noted in new pr: https://github.com/TestingResearchIllinois/idoft/pull/1497. Will update these README steps now since testing the adjusted scripts.

PR #1494, opened 2024-11-14
Update py-data.csv with RepoDeleted for tukio

Qi-Cui:


Qi-Cui:
> Great. Can you check `RepoDeleted` for all repos (in `py-data.csv`, `pr-data.csv`, and `gr-data.csv`)? You may look over the existing scripts to see if they help you.

Sure I will do that!

PR #1504, opened 2024-11-16
Update assigned flink test to DeveloperWontFix

mohitbadve:
The flink test was assigned to me in your_flaky_tests.csv, hence adding the related JIRA issue and closing it as DeveloperWontFix


darko-marinov:
I'm not sure how much it was `DeveloperWontFix` vs. `AStudentDidNotFollowUpAppropriately`. But fine to accept your change so we can move on.

PR #1529, opened 2024-11-21
Update deleted tests

zhihao11ui:


darko-marinov:
How did you find the commit? You can reply by email or here. Ideally, can you write a script (for some bonus "points") to automate the process?

zhihao11ui:
I did find it manually. I can definitely write a script to check which commit the file was deleted.

darko-marinov:
Please check the existing scripts to see what already exists. I'm accepting this PR with manual work but future cases should be done using some script (either previously written and potentially changed by you, or written by you from scratch).

PR #1553, opened 2024-12-02
Update two ID tests to DeveloperFixed

Jake-WangZhi:
These two ID tests in [djl](https://github.com/deepjavalibrary/djl) were fixed by the developer in [this commit](https://github.com/deepjavalibrary/djl/commit/7d197ba8c3c63bb6cb20b25c18b13ca12cd3233b), as they can no longer be detected as flaky tests.

I used the following command to find the most recent commit in the module where the tests were reported:

```
git log --oneline -- ./api
```

The last commit that touched the code was made after the flaky tests were reported, so I believe this is the commit that resolved the issue.

After running
```
./gradlew nondexTest --nondexRuns=10
```

The result is following
```
INFO: No Test Failed with this configuration.
INFO: *********
INFO: All tests pass without NonDex shuffling
INFO: ####################
INFO: Across all seeds:
INFO: Test results can be found at: 
INFO: file:///home/jakew4/djl/api/.nondex/HQ5kjB1HuYhogYODNehiqb0w1Ipnt50mEcmjPPVYKg=/test_results.html
INFO: [NonDex] The id of this run is: HQ5kjB1HuYhogYODNehiqb0w1Ipnt50mEcmjPPVYKg=
```


darko-marinov:
It's good that the tests don't fail after the commit. Do they fail right before the commit?

Jake-WangZhi:
> It's good that the tests don't fail after the commit. Do they fail right before the commit?

@darko-marinov It's actually pretty strange. I checked out several commits, and none of them caused these tests to fail. I even went back to a commit from two months ago, and the test still didn’t fail. However, this flaky test was first reported two weeks ago. I also reviewed the PR where this flaky test was introduced, but there was no description provided in the PR. Moreover, the commit associated with this flaky test only included changes to a YAML file, which shouldn't have caused the flakiness at all.

darko-marinov:
What PR introduced these tests? Hopefully the logs may still be available.

Jake-WangZhi:
> What PR introduced these tests? Hopefully the logs may still be available.

@darko-marinov Found it. here: https://github.com/TestingResearchIllinois/idoft/pull/1521

darko-marinov:
Closing because the tests couldn't be reproduced; we should start adding more info for reproducibility.

PR #1558, opened 2024-12-03
Update an ID test status to Accepted

Jake-WangZhi:
This PR has been `accepted`.

https://github.com/lets-blade/blade/pull/455


darko-marinov:
Great! Can you check if `lets-blade` has more `ID` tests (maybe after you finish that Gradle case)?

PR #1562, opened 2024-12-04
Add script to find which commit a test was deleted

zhihao11ui:


darko-marinov:
Did you run your script on a large number of repos to see what you get?

zhihao11ui:
I did not, this is intended for finding the commit of one test being deleted not for finding tests that were deleted. I have tested the script with the deleted tests that I submitted and it returned the correct commit

PR #1571, opened 2024-12-05
changed the status of  52 guava tests, fixed 11 new tests form swagger-core

xinying-7:
changed 52 guava test status to "DeveloperWontFix"


darko-marinov:
Please write a script to add `https://github.com/google/guava/pull/7425` in the last column (`Notes`) for all 52 tests that you changed.

xinying-7:
I've added notes to guava projects, and also 11 new test cases fixed. I've squashed all commits into one.

darko-marinov:
Please see email. @Ajitesh27 can you please check Guava cases.

Ajitesh27:
LGTM.
All the guava test cases are not for the same class that got Rejected, but the root cause for flakiness is the same which the developers decided not to fix.

PR #1580, opened 2024-12-06
Update an ID Test to MovedOrRenamed with script

Jake-WangZhi:
Learned and utilized the [auto-check-moved-or-rename script](https://github.com/TestingResearchIllinois/idoft/tree/main/auto-check-moved-or-rename) to identify the commit for this moved test.

The test were moved in [this commit](d305b4f153bccf676ef2d71033b11fb8a9de2090).

I put this test into the test.txt first and copy it and **check_moved_or_rename.sh** script to the root of the repo I wanna test and I ran this to find the commit.
```
./check_moved_or_rename.sh
```


darko-marinov:
How did you "I put this test into the `test.txt`"? Can you just grep for the test name? (Why) does the script require the input to be in a file named some way rather than being passed on stdin or as command-line arguments?

Jake-WangZhi:
> How did you "I put this test into the `test.txt`"? Can you just grep for the test name? (Why) does the script require the input to be in a file named some way rather than being passed on stdin or as command-line arguments?

@darko-marinov Yes, I believe we can just use grep to search for tests. However, this script expects an input text file containing all the tests to run, as mentioned in its README. From my understanding, the author likely wanted to keep the terminal cleaner by consolidating the test lines into a text file, making it easier to track what tests were executed. In the terminal, especially with multiple tests, the lines can wrap or get pushed to the next line, potentially creating a block of text that’s hard to read.

Jake-WangZhi:
@darko-marinov Sorry, I just realized I worked on the wrong repo—I was assigned the letter `l`, but somehow I thought it was `i`. Can we just close this PR?

Jake-WangZhi:
> @darko-marinov Sorry, I just realized I worked on the wrong repo—I was assigned the letter `l`, but somehow I thought it was `i`. Can we just close this PR?

PR #1582, opened 2024-12-06
1 dev won't fix

DaniellaHubble:
I updated my test from incubator-kie-drools as "developer won't fix" based on the interactions the developers had with a similar flaky test in the project. That pull request of that past pr is included in the notes.


darko-marinov:
@Ajitesh27 can you please check this case and also all other cases for this repo? If students don't want "point", you can open a PR to mark all these (so we don't revisit them in a few months).

PR #1586, opened 2024-12-06
12 dev won't fix tests

DaniellaHubble:
12 tests in my your_flaky_tests.csv from incubator-kie-drools (a repo in which developers have not wanted to fix flaky tests in the past)


darko-marinov:
How does this PR relate to #1582?

DaniellaHubble:
It is for the same thing as 1528 (just different tests). In retrospect I could have made it one PR, but I was looking at these tests after I had already made the prior PR for the other test.

Ajitesh27:
LGTM

PR #1609, opened 2025-09-19
[Test Status Update] Fixed and updated the status to Accepted for flaky test

Hussain-Safwan:
### What is the purpose of this PR

- The purpose of this PR is to resolve the unpredictability caused by the **flaky test**: `org.apache.dubbo.common.bytecode.WrapperTest.est_getMethodNames_ContainExtendsParentMethods`.
- The test passes/fails based on the ordering of the names of methods returned by`getMethodNames` methods from `Wrapper` and `ClassUtils` classes.

### Why the test fail

- The test asserts whether a certain array containing the names of the methods is available in the aforementioned classes.
- The problem, however, arises because the order in which the names would be returned is not guaranteed by the method used to retrieve the names.
- But `assertArrayEquals` expects even the order of the array elements to match, hence sometimes causing the failure. 

### How to reproduce the test failure

- The scenario may be reproduced by running the tests with **Nondex** tool:

-     `mvn -pl dubbo-common edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=org.apache.dubbo.common.bytecode.WrapperTest#test_getMethodNames_ContainExtendsParentMethods`

- Instructions on Nondex installation may be found [here](https://github.com/TestingResearchIllinois/NonDex).

### Expected results

The test should run successfully with Nondex instrumentation.

### Actual Results

- Following error result is obtained upon running the original code with Nondex: 
- `[ERROR] Failed to execute goal edu.illinois:nondex-maven-plugin:2.1.7:nondex (default-cli) on project dubbo-common: Unable to execute mojo: There are test failures.`
- Evidence that the test is indeed failing due to the difference in ordering can be observed in the following line: 
- `[ERROR]   WrapperTest.test_getMethodNames_ContainExtendsParentMethods:184 array contents differ at index [0], expected: <hello> but was: <world>`

### Description of fix

- The idea is to sort the array containing the names of methods before asserting.
- This can be performed using the `sort` method of `java.util` as in `Arrays.sort(methodNames)`.

### PR at the official repository

- A PR has been raised and **merged** in the official **[Apache/Dubbo](https://github.com/apache/dubbo)** repository in this regard
- The merged PR may be accessed here: https://github.com/apache/dubbo/pull/15683


pr-data.csv
```diff
@@ -1422,7 +1422,7 @@ https://github.com/apache/druid,50963edcae70150f13520b619f167512d951a71b,extensi
 https://github.com/apache/dubbo,737f7a7ea67832d7f17517326fb2491d0a086dd7,dubbo-filter/dubbo-filter-cache,org.apache.dubbo.cache.support.jcache.JCacheFactoryTest.testJCacheGetExpired,OD-Vic,MovedOrRenamed,,https://github.com/apache/dubbo/commit/acee3e2f03227894b5aa4d852de1a6a4bcfab60e
 https://github.com/apache/dubbo,5349c13a36d277a090e1dc68fbe7c3b46d78fc90,dubbo-common,org.apache.dubbo.common.beanutil.JavaBeanSerializeUtilTest.testDeserializeBean,OD-Vic,,,
 https://github.com/apache/dubbo,5349c13a36d277a090e1dc68fbe7c3b46d78fc90,dubbo-common,org.apache.dubbo.common.beanutil.JavaBeanSerializeUtilTest.testDeserialize_Array,OD-Vic,,,
-https://github.com/apache/dubbo,737f7a7ea67832d7f17517326fb2491d0a086dd7,dubbo-common,org.apache.dubbo.common.bytecode.WrapperTest.test_getMethodNames_ContainExtendsParentMethods,ID,Rejected,https://github.com/apache/dubbo/pull/6978,
+https://github.com/apache/dubbo,737f7a7ea67832d7f17517326fb2491d0a086dd7,dubbo-common,org.apache.dubbo.common.bytecode.WrapperTest.test_getMethodNames_ContainExtendsParentMethods,ID,Accepted,https://github.com/apache/dubbo/pull/15683,
```
darko-marinov, line 1425: Please add `Rejected https://github.com/apache/dubbo/pull/6978` into `Notes` (the last column).

Hussain-Safwan: @darko-marinov Hi, I've added the rejected PR to the `Notes` column.

Hussain-Safwan:
> Please squash the two commits. We don't need two commits to change one line.

I have squashed the commits.

PR #1611, opened 2025-09-29
Open PR in apache/nifi, apache/tinkerpop, and apache/hive

KangweiZhu:
* One Accepted: `apache/nifi`
* One Opened: `apache/hive`


KangweiZhu:
* Four Opened: apache/tinkerpop

PR #1614, opened 2025-10-01
Open PR in openhab/openhab-core

EdwinIngJ:
- Opened PR to fix 11 tests in `openhab/openhab-core`


darko-marinov:
Your PR for openhab is quite big; maybe split such big PRs into smaller ones in the future -- you could have opened one PR for one class and (maybe later) another PR for the other class.

PR #1615, opened 2025-10-01
Open PR & Change of Status in vertx.io

cperkkk:
Open 1 PR in `eclipse-vertx/vert.x`, fixing 1 flaky test


darko-marinov:
Rather than calling the test flaky, it's better to describe the root cause.

PR #1616, opened 2025-10-01
open PR in adobe/aem-core-wcm-components

kabo87777:
Open 1 PR in `adobe/aem-core-wcm-components`, fixing 2 tests


darko-marinov:
Rather than calling the test flaky, it's better to describe the root cause. Also, you could have opened two PRs, one for each test (the first change seems to be much smaller than the second change). Further, maybe the two commits could have been squashed (or moved into different PRs). Last but not least, there's an unnecessary end-of-line change at the end of the file.

PR #1617, opened 2025-10-01
Update pr-data.csv

cperkkk:
Doesn't seem to be valid, I run nondex twice, for a few minutes on root and singular test, no result.

image

Test also doesn't look like an ID flaky test

image


darko-marinov:
1. Please do not upload pictures but put a log file on your VM and let me know *privately* (no need to share your NetID or VM ID here) the VM ID and the absolute path to your log file.
2. Run NonDex on the entire project with `-DnondesRuns=10`.

PR #1618, opened 2025-10-01
Open PR in spring-projects/spring-retry

chihyu0917:
Open 1 PR in `spring-projects/spring-retry`, fixing 3 tests


darko-marinov:
I'm accepting this PR but please remove the end-of-file change from the real PR -- that PR is already fairly big as is.

PR #1620, opened 2025-10-01
Opened PR for google/guava

annhchen89:
Opened 1 PR in ```google/guava```


darko-marinov:
From the initial discussion, your real PR may be rejected for reasons similar to another. If it does happen, please find for each and every test in Guava if it fails only in FULL mode or also in ONE mode. You could then mark those that fail only in FULL as DevelopersWontFix and add a note at the end.

PR #1633, opened 2025-10-02
Open PR for 3 flaky tests from `apache/servicecomb-java-chassis`

pem70:
PR: https://github.com/apache/servicecomb-java-chassis/pull/4967
This PR fixes 3 flaky tests


darko-marinov:
Please do not list the real PR link in the IDoFT PR description. We already have the link in `pr-data.csv`, and we need not add noise with "mentioned in", as I showed in the class.

PR #1654, opened 2025-10-03
added 1944

18chanp1:
found the commit


darko-marinov:
Please put on your VM logs showing that (1) the test was `ID` in `551615fa13f65ae996bae9c1bacff189539b6557^` (before changes) and (2) not `ID` (with `-DnondexRuns=100` or `=10` if too slow, ideally even with the entire test class). Email me privately your VM ID and absolute path to the log files.

darko-marinov:
Please rebase your changes on the latest `main`.

18chanp1:
Upon further testing, I have discovered it still exists. It is hard to detect and requires a large number of runs. I have updated this to a "opened" pr with a "fix".

PR #1666, opened 2025-10-06
open PR in apache/iotdb

kabo87777:
Open 1 PR in `apache/iotdb`, fixing 5 tests
## Why Single PR for All 5 Tests
- **Shared infrastructure**: All tests use the same `checkVisitorResult` helper methods that needed modification
- **Atomic fix**: The core solution (Set-based comparison) addresses all 5 tests simultaneously


darko-marinov:
I'm accepting this PR here, but the real PR looks somewhat big (+71, -67 lines).

PR #1707, opened 2025-10-14
Detected and fix 5 flaky test

cperkkk:
Detected and fix 5 flaky test in ```FasterXML/jackson-databind```


darko-marinov:
Can you please see why the checker fails? The message in https://github.com/TestingResearchIllinois/idoft/actions/runs/18511703513/job/52753860988?pr=1707 is confusing, but I think the problem is ordering (`ser` should be before `struct`).

PR #1721, opened 2025-10-17
Deleted entry in apache/shardingsphere-elasticjob

bee0511:


pr-data.csv
```diff
@@ -3500,7 +3500,7 @@ https://github.com/apache/shardingsphere-elasticjob,b022898ef1b8c984e17efb2a422e
...
 https://github.com/apache/shardingsphere-elasticjob,b022898ef1b8c984e17efb2a422ee45f6b13e46e,elastic-job-lite-core,io.elasticjob.lite.event.rdb.JobEventRdbStorageTest.assertAddJobStatusTraceEventWhenFailoverWithTaskStagingState,NIO,Accepted,https://github.com/apache/shardingsphere-elasticjob/pull/1838,
-https://github.com/apache/shardingsphere-elasticjob,b022898ef1b8c984e17efb2a422ee45f6b13e46e,elastic-job-lite-core,io.elasticjob.lite.executor.type.SimpleJobExecutorTest.assertExecuteWhenRunOnceAndThrowExceptionForMultipleShardingItems,ID,,,
+https://github.com/apache/shardingsphere-elasticjob,b022898ef1b8c984e17efb2a422ee45f6b13e46e,elastic-job-lite-core,io.elasticjob.lite.executor.type.SimpleJobExecutorTest.assertExecuteWhenRunOnceAndThrowExceptionForMultipleShardingItems,ID,MovedOrRenamed,,https://github.com/apache/shardingsphere-elasticjob/commit/556c6c76e2311a411e449b02f3c4b655268f978c
```
darko-marinov: Can you add a new line for where the test is moved? Did you check if it was still `ID` after being moved?

bee0511: > Yes, if the test was fixed before being renamed, then we should find the fix not the rename.
> 
> Did you also check if it's `ID` in the latest commit (in the new location)?

I found that the test `assertExecuteWhenRunOnceAndThrowExceptionForMultipleShardingItems` was deleted in `ca4050db6277a1ebfefd80e2d9bb75fd45d1ed11`.

The commit `556c6c76e2311a411e449b02f3c4b655268f978c` moved some other class with the same file name, but the whole test functions within that class are different.

You can use the command `git show ca4050db6277a1ebfefd80e2d9bb75fd45d1ed11 --stat -- "*SimpleJobExecutorTest.java"` to double check that they did delete the test.

darko-marinov: Do you want to change `MovedOrRenamed` to `Deleted`? I may still ask if the test was `ID` before being deleted. :)

darko-marinov: Thanks for carefully checking. I'll accept as `Deleted` after you squash your two commits into one.

PR #1728, opened 2025-10-19
Update: Developer fixed tests in swagger-jaxrs2

Fr4nk03:
tests in module swagger-jaxrs2 are no longer flaky

git bisect is used to detect the fixed commit:
https://github.com/swagger-api/swagger-core/commit/0d067d03b6ef78a78efe3e4f957f977254c72030


darko-marinov:
Please see my email about the specific hunk in that commit.

PR #1756, opened 2025-10-23
opened prs in hadoop.yarn.util.resource.TestResources

18chanp1:


darko-marinov:
The checker failed.

darko-marinov:
Why does the real PR need to clean in both setup and teardown? That PR description lists 11 tests but only 10 are here.

18chanp1:
> Why does the real PR need to clean in both setup and teardown? That PR description lists 11 tests but only 10 are here.

unsetResourceType deletes a file structure, so cleaning during teardown prevents this test from polluting other tests, and cleaning during setup prevents this test from being a victim of any other test in the future. 

I copied the next line from the spreadsheet in error, I have edited the real PR to remove the 11th test.

PR #1761, opened 2025-10-23
Adding 1 deleted test for apache/dubbo

Anshul-creator:
The following test no longer exists on active Dubbo branches. The entire dubbo-monitor module (which contained DubboMonitorTest) was removed in 3.3 and parts were folded into the dubbo-metrics modules. The test itself was not renamed or moved—it was deleted with the module.

I created a script to put all the logs into a file for POW. Will send the details in email


darko-marinov:
Accepting although you could have put the entire commit SHA.

PR #1772, opened 2025-10-25
GraphChiPageRankOperatorTest.testExecution moved to a new folder

minizhiren:
After cross-checking with the upstream repository history, we found that the earlier assumption about this test being moved was based on outdated information.

In reality, GraphChiPageRankOperatorTest.testExecution had already been deleted upstream prior to the SHA where the issue was observed.
Therefore, the previous interpretation (“moved to a new folder”) was inaccurate.

This PR now reflects the correct state by removing the obsolete test, aligning with the upstream codebase.


pr-data.csv
```diff
@@ -2884,7 +2884,7 @@ https://github.com/apache/incubator-shenyu,3777209fc76800175df3962213ca1febe57ed
...
 https://github.com/apache/incubator-streampipes,7ba9213012b72ea70edc6848c18ba6d90ff2c534,streampipes-connect,org.apache.streampipes.connect.adapter.format.json.xml.XmlTest.parseEventParkingFaciltyStatus,ID,Accepted,https://github.com/apache/incubator-streampipes/pull/59,
 https://github.com/apache/incubator-uniffle,6fb2a9a63132ab9abbde2f1f9240f9caf2f5d0f0,common,org.apache.uniffle.common.rpc.GrpcServerTest.testGrpcExecutorPool,TD,,,
-https://github.com/apache/incubator-wayang,b7b6c1ebf65b58f5c03ac2b9fd4c1846ea877d01,wayang-platforms.wayang-graphch,org.apache.wayang.graphchi.operators.GraphChiPageRankOperatorTest.testExecution,ID,,,
+https://github.com/apache/incubator-wayang,b7b6c1ebf65b58f5c03ac2b9fd4c1846ea877d01,wayang-platforms.wayang-graphch,org.apache.wayang.graphchi.operators.GraphChiPageRankOperatorTest.testExecution,ID,MovedOrRenamed,,https://github.com/apache/incubator-wayang/tree/8e8897900921e52283c1c6093e986e51a3737393
```
darko-marinov: Change the link to have `/commit/` instead of `/tree/`. If this commit changed the test, can you check if the test was `ID` and add a new line with the new name/location?

minizhiren:
sorry! That button looks like "refresh'. I just try to'refres' for something new

darko-marinov:
Okay, it's good that you found the test was deleted. I could ask whether it was still `ID` right before being deleted, but let me leave that question for another time and accept this change now.

PR #1777, opened 2025-10-28
Added one Open PR to apache/dubbo

Anshul-creator:
Created and opened a PR in apache/dubbo for the test `ExtensionLoaderTest.testInjectExtension`


darko-marinov:
Seems like a challenging case. Presumably the test doesn't fail intermittently but deterministically based on the order.

PR #1793, opened 2025-10-31
Developer fixed test in Activiti/Activiti

Leo-Fish:
the module Activiti and Activiti-core is merged, so all the tests get moved into Activiti-core and its sub-modules. Still working to check if all of them is moved.


darko-marinov:
Was this resolved on some tagged/released commit?

Leo-Fish:
checking on the format error

darko-marinov:
Nice workaround for the format check. Accepting!

PR #1797, opened 2025-11-01
updated opened prs to rejected as per previous pr comment and email

lawrencewang49:


darko-marinov:
Move on to other projects but be more careful with the PRs. It's not a problem to have a few PRs rejected, but you don't want many rejected.

PR #1810, opened 2025-11-05
apache/shardingsphere-elasticjob 1 deleted test

bee0511:
one more deleted test :)


darko-marinov:
Can you manually check some parts of your AI-generated script and then use it to improve that `auto-check-deletion`. For example, does your script treat the Git history as a list (you mention "range"), or does it treat it as a graph (as it should). Again, you need *not* handle the most general cases. It suffices to be somewhat better than the current script. I don't even want any potentially "perfect" scripts that students could just run without learning what goes on.

bee0511:
> Can you manually check some parts of your AI-generated script and then use it to improve that `auto-check-deletion`. For example, does your script treat the Git history as a list (you mention "range"), or does it treat it as a graph (as it should). Again, you need _not_ handle the most general cases. It suffices to be somewhat better than the current script. I don't even want any potentially "perfect" scripts that students could just run without learning what goes on.

I looked at both the auto-check-deletion script and my own version. Right now, both treat the Git history like a list, not a graph. I can improve it to be more graph-aware by checking all parent commits by running something like `git show -s --pretty=%P` for each commit to compare the method’s presence in all parent trees. I will improve the script in https://github.com/TestingResearchIllinois/idoft/tree/main/auto-check-deletion with these changes and open a PR later.

PR #1845, opened 2025-11-13
MovedOrRenamed Tests in Apache/Nifi

ghost:
Details for the tests can be found at the VM: fa25-cs527-092

Script ran: /home/siddhij2/project/nifi/check_moved_or_rename.sh
Output File: /home/siddhij2/project/nifi/output.txt


darko-marinov:
Please squash your three commits into one.

Check if the test was `ID` right after being moved.

KangweiZhu:
Just checked with Siddhi, this should be an individual PR (we created another one earlier today about changing the tests' status from Opened to Accepted in TinkerPop), but we mistakenly opened two PRs from the same branch in forked repo. We will avoid this in the future.

KangweiZhu:
I ran NonDex on that module and on specific tests, they passed without any flaky behavior.

SiddhiJhunjhunwala-UIUC:
LdapUserGroupProviderTest in nifi-ldap-iaa-providers-bundle is deterministic and not flaky. 
No commit or JIRA exists that fixes flakiness for this test class. 
PR #6946 (Feb 13, 2023) only migrated the module to JUnit 5 and did not change test behavior. 
Therefore, there is no fix commit; the test is inherently stable.

darko-marinov:
@SiddhiJhunjhunwala-UIUC can you please change `MovedOrRenamed` to `DeveloperFixed`?

PR #1872, opened 2025-11-19
Added 1 opened test, fixed a few entries

Anshul-creator:
Added entry for a test I opened a PR last week and forgot to add to idoft.
Fixed a few entries in the pr-data.csv file. I had the module path incorrect for the entries I made in previous 3 commits since I didn't include the parent (for example, there is no folder dubbo-config-api but rather the folder is a subfolder in the directory dubbo-config hence correct entry is dubbo-config/dubbo-config-api)


darko-marinov:
Thanks for correcting the module names. I should more carefully check what I accept. Do you want to extend https://github.com/TestingResearchIllinois/idoft/tree/main/format_checker to check that module names exist in the reported SHAs?

PR #1085, opened 2023-11-26
Detected New Flaky Test in Pinot-query-planner

Nemesis123925:
Reported detected flaky tests in Pinot-query-planner

[Campuswire Post](https://campuswire.com/c/G7A0E96FD/feed/681)

I replaced `\n` and `\r` in the original test cases by `space`

PR #1900, opened 2025-11-30
Found 1 DeveloperFixed Pinot Test

LucasEby:
One of the previously discovered ID tests in Apache Pinot was fixed by the developers. I was able to track down the commits from the last time it failed and when it was fixed. I have included instructions below to reproduce my results as well as the locations of the output files on the VM as proof of the results.

## Output File Locations on VM
Failing output: `014:~/pinotDeveloperFixed/a02253d671d7781acc16b81628e51bd11a3005a0Output.txt`

Passing output: `014:~/pinotDeveloperFixed/782b6979360b6ec28c869bbf8b7ea59e1548a3efOutput.txt`

### How to Reproduce the final FAILING ID Test
1. Checkout the SHA of the last FAILED commit:
`git checkout a02253d671d7781acc16b81628e51bd11a3005a0`

2. Delete this line from the pom.xml:
-javaagent:${settings.localRepository}/org/mockito/mockito-core/${mockito-core.version}/mockito-core-${mockito-core.version}.jar

3. Compile it:
`./mvnw clean install -DskipTests -Pbin-dist -Pbuild-shaded-jar -Dlicense.skip   -Drat.skip -Dcheckstyle.skip`

4. Run the ID test with nondex. We are expecting nondex to FAIL after running the below command. Less than 10 iterations was found to be sufficient to generate the error, so we ran 100 just to be on the safe side.
`mvn -pl pinot-query-planner -Dtest=org.apache.pinot.query.queries.ResourceBasedQueryPlansTest#testQueryExplainPlansWithExceptions -DnondexRuns=100 -DreuseForks=false  edu.illinois:nondex-maven-plugin:2.2.1:nondex -Dcheckstyle.skip=true -DnondexRunsWithoutShuffling=0`

### How to Reproduce Fixed ID Test:
1. Checkout commit that FIXED the ID Test: 
`git checkout 782b6979360b6ec28c869bbf8b7ea59e1548a3ef`

2. Delete this line from the pom.xml:
-javaagent:${settings.localRepository}/org/mockito/mockito-core/${mockito-core.version}/mockito-core-${mockito-core.version}.jar

3. Compile it:
`./mvnw clean install -DskipTests -Pbin-dist -Pbuild-shaded-jar -Dlicense.skip   -Drat.skip -Dcheckstyle.skip`

4. Run the ID test with nondex. We are expecting nondex to PASS after running the below command. Less than 10 iterations was found to be sufficient to generate the error, so we ran 100 just to be on the safe side to ensure that it was fixed.
`mvn -pl pinot-query-planner -Dtest=org.apache.pinot.query.queries.ResourceBasedQueryPlansTest#testQueryExplainPlansWithExceptions -DnondexRuns=100 -DreuseForks=false  edu.illinois:nondex-maven-plugin:2.2.1:nondex -Dcheckstyle.skip=true -DnondexRunsWithoutShuffling=0`


pr-data.csv
```diff
@@ -3495,7 +3495,7 @@ https://github.com/apache/pinot,ecf41be2ecd007853c2db19e1c6a038cf356cb9e,pinot-c
 https://github.com/apache/pinot,ecf41be2ecd007853c2db19e1c6a038cf356cb9e,pinot-core,org.apache.pinot.queries.NullEnabledQueriesTest.testQueriesWithNoDictDoubleColumn,ID,,,
 https://github.com/apache/pinot,ecf41be2ecd007853c2db19e1c6a038cf356cb9e,pinot-core,org.apache.pinot.queries.NullEnabledQueriesTest.testQueriesWithNoDictFloatColumn,ID,DeveloperFixed,,https://github.com/apache/pinot/commit/db0f3dcb73bd338d4b754329679e5f3e4d6bf7e0
 https://github.com/apache/pinot,f79b618141e07c140663791959d96956ad91d811,pinot-query-planner,org.apache.pinot.query.queries.ResourceBasedQueryPlansTest.testQueryExplainPlansAndQueryPlanConversion,ID,,,fails 21/375 cases
-https://github.com/apache/pinot,f79b618141e07c140663791959d96956ad91d811,pinot-query-planner,org.apache.pinot.query.queries.ResourceBasedQueryPlansTest.testQueryExplainPlansWithExceptions,ID,,,fails 1/38 cases
+https://github.com/apache/pinot,f79b618141e07c140663791959d96956ad91d811,pinot-query-planner,org.apache.pinot.query.queries.ResourceBasedQueryPlansTest.testQueryExplainPlansWithExceptions,ID,DeveloperFixed,,https://github.com/apache/pinot/pull/16109
```
darko-marinov: Great finding. What was `fails 1/38 cases`; why delete it?

darko-marinov: As part of "Git(Hub) archaeology", can you fin who left the note and why?

For the failing check, the piece of information in `Notes` should be separated by `;` not `,`.

darko-marinov: How did you find the original PR? :) The summary is good. Just change that last `,` to `;`, and I can accept your PR.

LucasEby: The format checker doesn't like it if there are spaces in the note for a DeveloperFixed entry. Most of the confusion was because I found a line in the csv file which is DeveloperFixed and does contain spaces. I'm guessing that this line was added prior to the current format checker being incorporated which explains why I wasn't able to get any combination of entries to work with spaces (I even tried their exact note and it didn't work).

darko-marinov: > Git bisect gave me the PR that was associated with the commit

Two questions here:

1. How would Git bisect give the PR (it gives just the commit)?

2. Are we talking about the same PR/commit? I was talking about #1085, which I think you found in another way, not using Git bisect.

PR #1928, opened 2025-12-05
Change test status from Opened to Accepted

KangweiZhu:
These tests were accepted three years ago. I guess they forgot to update tests' status because the semester had ended.


darko-marinov:
Can you please run the existing scripts in IDoFT to check the status of all `Opened` PRs?

darko-marinov:
You can start from that script but may need to change it. Also, you need not run on all (currently 1,305) rows with `,Opened,` but run on some of them.

PR #1975, opened 2025-12-16
Opened some PRs in OpenSearch

liuguoqingfz:


liuguoqingfz:
> I can accept this PR here now that the checker is passing, but your real PRs are failing some checks, e.g., for `CHANGELOG.md`. Please address those checks; otherwise, your real PRs are likely to be (considered) rejected.

I believe the adding fix for tests to CHANGELOG are not required according to their [documentation](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#which-changes-require-a-changelog-entry). Also, for the other gradle checks, often time  it fails because it has to run the entire CI again and some other flaky tests failed causing the checks to fail, it doesn't necessarily mean that the tests that I was changing are causing the failure. It 's just because of the nature of the large amount of flaky tests there are in the repo.

PR #1980, opened 2025-12-17
Added a PR

liuguoqingfz:


gr-data.csv
```diff
@@ -719,6 +720,7 @@ https://github.com/opensearch-project/OpenSearch,6966267722419e7a478305313c0ddaa
...
 https://github.com/opensearch-project/OpenSearch,3506973761972df391cabef40b9858e6c05239f0,OpenSearch/modules/mapper-extras/src/test/java/org/opensearch/index/mapper,org.opensearch.index.mapper.ScaledFloatFieldMapperTests.testScaledFloatWithStarTree,NOD,Opened,https://github.com/opensearch-project/OpenSearch/pull/20254,
 https://github.com/opensearch-project/OpenSearch,bd98dfb3e5fee6bd8c88667841088149dff5da33,OpenSearch/server/src/internalClusterTest/java/org/opensearch/index/shard,org.opensearch.index.shard.IndexShardIT.testIndexCanChangeCustomDataPath,NOD,,,
+https://github.com/opensearch-project/OpenSearch,c3533a7c2a3c17d2dfadb270b60d1c8006b687dd,OpenSearch/server/src/internalClusterTest/java/org/opensearch/indices/replication,org.opensearch.indices.replication.MergedSegmentWarmerIT.testPrimaryStopped_ReplicaPromoted,NOD,Deleted,,https://github.com/opensearch-project/OpenSearch/issues/19042
```
darko-marinov: What commit deleted this test? Your not has some issue that talks about several tests.

liuguoqingfz: What do you mean by "Your not has some issue that talks about several tests."? Do you need me to find out which SHA was this test deleted?

liuguoqingfz: The method was actually in the parent class, I couldn't find the method in the child class so I thought it was deleted. I've opened a PR for the fix in the parent class, so I changed Deleted to Opened, and there was another test that has the same situation as this one but I don't think I have enough time to fix that so I just change that Deleted to Blank

